PR #24, opened 2018-06-11
Feature export ary

bewakes:


ary/models.py
```diff
@@ -279,3 +283,185 @@ def can_modify(self, user):
             (self.lead and self.lead.can_modify(user)) or
             (self.lead_group and self.lead_group.can_modify(user))
         )
+
+    def _load_schema(self, FieldClass, GroupClass=None):
```
bibekdahal: GroupClass cannot be None based on the function definition.

A more suitable name may be `get_schema` or `create_schema` or `generate_schema`: this doesn't load schema into the object but rather generate a new schema.

bibekdahal: `FieldClass` is also never used and should be removed. So the function definition can be something like:

```
def create_schema_for_group(self, GroupClass):
```

ary/models.py
```diff
@@ -279,3 +283,185 @@ def can_modify(self, user):
...
+                    'options': {
+                        x['key']: x['title'] for x in field.get_options()
+                    }
+                }
+                for field in group.fields.all()
+            ] for group in groups
+        }
+        return schema
+
+    def _data_from_schema(self, schema, raw_data):
```
bibekdahal: I think we should write all function names (especially functions that get or set values) as an imperative name, that starts with a verb. `get_data_from_schema` would be a better name.

bibekdahal: Perhaps this method can be `static` or a separate utility function? It doesn't seem to use any data of `self`.

ary/models.py
```diff
@@ -279,3 +283,185 @@ def can_modify(self, user):
...
+        return {
+            underscore_to_title(k):
+                v if not isinstance(v, list)
+                else [mapping[k](y) for y in v]
+            for k, v in methodology_raw.items()
+        }
+
+    def get_summary_json(self):
+        # functions to get exact value of an entry in summary group
+        def identity(x):
```
bibekdahal: Duplicate `identity` function that is already defined in *utils/common.py*.

ary/models.py
```diff
@@ -279,3 +283,185 @@ def can_modify(self, user):
...
+
+        # Formatting of underscored keywords, by default is upper case as given
+        # by default_format() function below
+        formatting = {
+            'priority_sector': lambda x: 'Most Unmet Needs Sectors',
+            'affected_location': lambda x: 'Settings Facing Most Humanitarian Issues'  # noqa
+        }
+
+        default_format = underscore_to_title   # function
+
```
bibekdahal: Useless function definition as `underscote_to_title` is already defined.

export/views.py
```diff
@@ -5,13 +5,17 @@
...
 
 from export.serializers import ExportSerializer
 from export.models import Export
 from project.models import Project
 
-from export.tasks_entries import export_entries
+from export.tasks import export_entries, export_assessment
+
+import logging
+logger = logging.getLogger('django')
```
bibekdahal: Is this used?

bibekdahal (changes requested):

ary/models.py
```diff
@@ -279,3 +283,185 @@ def can_modify(self, user):
...
+        def default_format(x):
```
bibekdahal: Useless function definition as `underscote_to_title` is already defined.

bibekdahal:
For future: always write commit message as an imperative sentence and start with a capital letter.

References:
https://github.com/the-deep/deeper/blob/develop/docs/git.md#commit-messages
https://chris.beams.io/posts/git-commit/

ary/models.py
bewakes: sahi ho dahal.

export/views.py
bewakes: It was used, later I removed logs and forget this one. My bad.

ary/models.py
bewakes: Roger That

ary/models.py
bewakes: Copy That.

bibekdahal (approved):

PR #38, opened 2018-06-29
Add Sentry to Celery

thenav56:
Capture Exception.
May require [manual capture](https://github.com/the-deep/server/blob/develop/deep/exception_handler.py#L28#L35) on custom exception handling.


bibekdahal:
Moved the remaining task to new issue: #45

PR #42, opened 2018-07-04
Feature new connectors

bewakes:


eoglethorpe:
@bewakes @bibekdahal do we have any failsafes here incase the website structures change and this code no longer works?

PR #43, opened 2018-07-05
Update email logic

thenav56:
* Add Unsubscribe in base email template
* Add Token for accepting request without login


project/views.py
```diff
@@ -130,26 +137,37 @@ def join_project(self, request, pk=None, version=None):
...
             # send the join request emails in a celery
             # background task.
             # This makes sure that the response is returned
             # while the emails are being sent in the background.
             def send_mail():
-                send_project_join_request_emails.delay(
-                    request.user.id,
-                    project.id,
-                )
+                send_project_join_request_emails(join_request.id)
```
bibekdahal: `send_project_join_request_emails.delay(join_request.id)` should be used for triggering celery task

bibekdahal (commented):
Also, add support for "reject" and "accept as admin" in emails.

project/views.py
thenav56: Fixed!

PR #67, opened 2018-08-17
Fix Email template viewer

thenav56:
* Also change nightly deepl domain to `https://deepl-nightly.thedeep.io`

### DIFF
--- a/deep/views.py
+++ b/deep/views.py
@@ -6,6 +6,9 @@ from django.views.generic import View
 from django.conf import settings
 from django.template.response import TemplateResponse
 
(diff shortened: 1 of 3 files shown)


deep/views.py
```diff
@@ -30,26 +33,33 @@ def get(self, request, exception):
                        code=status.HTTP_404_NOT_FOUND)
 
 
+def get_basic_email_context():
+    user = User.objects.get(pk=1)
```
bewakes, line 37: It's okay, but better to use safe code here, e.g: user.objects.last() or user.objects.first().

PR #69, opened 2018-08-17
Add api for seraching users and user groups

bewakes:


user_group/views.py
```diff
@@ -66,3 +68,30 @@ def finalize_response(self, request, response, *args, **kwargs):
...
         return GroupMembership.get_for(self.request.user)
+
+
+class UserGroupUserSearchViewSet(viewsets.ViewSet):
+    permission_classes = [permissions.IsAuthenticated]
+
+    def list(self, request, version=None):
+        query = request.query_params.get('query', '').strip().lower()
+        if not query:
+            return response.Response(
```
bibekdahal: Instead of returning response, raise a `exceptions.ValidatorError` so that the error format is actually handled by `exception_handler`.

user_group/views.py
```diff
@@ -66,3 +68,30 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+            Q(last_name__icontains=query) |
+            Q(email__icontains=query)
+        ).\
+            annotate(title=F('username'), type=Value('user', CharField())).\
+            values('id', 'title', 'type')
+        user_groups = UserGroup.objects.filter(title__icontains=query).\
+            annotate(type=Value('user_group', CharField())).\
+            values('id', 'title', 'type')
+        items = list(users) + list(user_groups)
+        return response.Response(
```
bibekdahal: For consistent data structure expected by client, the result should be in the paginated format. Use `self.get_paginated_response(items)`.

user_group/views.py
```diff
@@ -66,3 +68,30 @@ def finalize_response(self, request, response, *args, **kwargs):
 
     def get_queryset(self):
         return GroupMembership.get_for(self.request.user)
+
+
+class UserGroupUserSearchViewSet(viewsets.ViewSet):
+    permission_classes = [permissions.IsAuthenticated]
+
+    def list(self, request, version=None):
+        query = request.query_params.get('query', '').strip().lower()
```
bibekdahal: Instead of `query` let's use `search` as it is used by DRF by default in other APIs.

user_group/views.py
bibekdahal: Actually if we union the `users` and `user_groups` into single query set, we can use serializer and let django handle the pagination and response format.

bibekdahal:
I have made the requested changes myself due to complexity. @bewakes  Please take a look at the changes for future references. :stuck_out_tongue:

PR #71, opened 2018-08-22
Fix from feedback

thenav56:
Ref: https://docs.google.com/document/d/1tn-fQy2WzmRukR8Q6bLggZ0-DpJM776R31Ae7nLdnXc/edit

### DIFF
--- a/templates/user/unsubscribe_email__confirm.html
+++ b/templates/user/unsubscribe_email__confirm.html
@@ -5,12 +5,11 @@
 
     {% if success %}
         <p>
-            Successfully Unsubscribed
+            Successfully unsubscribed
         </p>
     {% else %}
         <p>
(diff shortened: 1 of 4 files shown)


templates/user/unsubscribe_email__confirm.html
```diff
@@ -5,12 +5,11 @@
...
     {% if success %}
         <p>
-            Successfully Unsubscribed
+            Successfully unsubscribed
         </p>
     {% else %}
         <p>
-            Unsubscribe link is invalid!
-            you must be already unsubscribed
+            Unsubscribe link is invalid!  you must be already unsubscribed
```
bibekdahal, line 12: @eoglethorpe Can you provide the text for when the user clicks on the unsubscribe link after he has already unsubscribed once?

PR #74, opened 2018-08-24
Feature add project to entities

bewakes:


entry/serializers.py
```diff
@@ -76,22 +77,33 @@ class Meta:
...
         model = Entry
-        fields = ('id', 'lead', 'analysis_framework',
+        fields = ('id', 'lead', 'analysis_framework', 'project',
                   'entry_type', 'excerpt', 'image', 'information_date',
                   'attributes', 'order', 'client_id',
                   'created_at', 'created_by', 'modified_at', 'modified_by',
                   'version_id')
 
+    def create(self, data):
+        project = data['project']
```
bibekdahal: This function seems pointless at the moment. Following should be done instead:

* If `data.get('project')` is None, set `data['project']` to `data['lead'].project`
* Do the same for assessment
* Write test when project is not provided

lead/models.py
```diff
@@ -166,12 +167,6 @@ def get_for(user):
     def can_get(self, user):
```
bibekdahal: Not sure but `can_get` should probably be removed from here.

bibekdahal: Also remove old `get_for` method. And remove these from Entry and Assessment too.

project/models.py
```diff
@@ -172,7 +187,7 @@ def get_modifiable_for(user):
         return Project.get_annotated().filter(
             projectmembership__in=ProjectMembership.objects.filter(
                 member=user,
-                role='admin',
+                role=ProjectRole.get_admin_role(),
```
bibekdahal: Should use setup_permissions instead of testing role = admin.

PR #94, opened 2018-09-26
Add project-usergroup m2m through model

bewakes:


frozenhelium:
3 failed tests @bewakes

PR #133, opened 2018-11-30
Feature notification model

frozenhelium:


deep/settings.py
```diff
@@ -83,7 +83,7 @@
     'user_resource',
     'tabular',
     'commons',
-
+    'notification',
```
thenav56, line 86: Add a new line after `notification`. 
right now the block for the install_app is as
```
# Django apps
# lib apps
# Deep apps
# Misc apps
```

PR #137, opened 2018-12-02
Fix notification api and add testcases

bewakes:


notification/serializers.py
```diff
@@ -2,22 +2,39 @@
...
-from project.serializers import (
-    ProjectJoinRequestSerializer,
-)
-from project.models import (
-    ProjectJoinRequest,
-)
 from .models import (Notification)
 
 
+class NotificationStatusSerializer(serializers.Serializer):
```
bibekdahal: Isn't this the same as `NotificationSerializer`? As long as other fields are `read_only` in NotificationSerializer, I think that can be used instead of `NotificationStatusSerializer`. Also, we do not need validation functions if we make the serializer a ModelSerializer.

notification/serializers.py
```diff
@@ -2,22 +2,39 @@
...
     data = serializers.SerializerMethodField()
 
     class Meta:
         model = Notification
         fields = ('__all__')
+        read_only_fields = (
+            'data', 'receiver', 'project', 'notification_type'
+        )
 
     def get_data(self, notification):
```
bibekdahal, line 32: I don't think we need this method at all now.

notification/tests/test_apis.py
```diff
@@ -0,0 +1,111 @@
...
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+class TestNotificationAPIs(TestCase):
+    def setUp(self):
+        super().setUp()
+        # Clean up notifications
+        Notification.objects.all().delete()
```
bibekdahal: Do we need this? Isn't database in a cleaned up state during setUp by default?

notification/tests/test_apis.py
```diff
@@ -0,0 +1,111 @@
+from deep.tests import TestCase
+
+from user.models import User
+from notification.models import Notification
+from project.models import ProjectJoinRequest, Project
+
+import logging
+logger = logging.getLogger(__name__)
```
bibekdahal: Is it necessary to import these?

notification/tests/test_notification.py
```diff
@@ -0,0 +1,131 @@
+from deep.tests import TestCase
+
+from user.models import User
+from notification.models import Notification
+from project.models import ProjectJoinRequest, Project
+
+import logging
+logger = logging.getLogger(__name__)
```
bibekdahal: See my comment on `test_apis`

notification/tests/test_notification.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class TestNotification(TestCase):
+    """Unit test for Notification"""
+    def setUp(self):
+        super().setUp()
+        # Clear all notifications
+        Notification.objects.all().delete()
```
bibekdahal: See my comment on `test_apis`

notification/views.py
```diff
@@ -29,22 +35,23 @@ def filter_queryset(self, queryset):
...
-        return result
+    @list_route(permission_classes=[permissions.IsAuthenticated],
+                methods=['put'],
+                url_path='status')
+    def status_update(self, request, version=None):
+        serializer = NotificationStatusSerializer(data=request.data, many=True)
+        valid = serializer.is_valid()
+        if not valid:
+            raise exceptions.ValidationError(serializer.errors)
+        for item in serializer.data:
```
bibekdahal: If this was NotificationSerailizer, then I think the a simple `serializer.save()` would do instead of this for loop.

notification/views.py
```diff
@@ -29,22 +35,23 @@ def filter_queryset(self, queryset):
...
-
-        if not getattr(request, 'child_route', False):
-            queryset = self.paginate_queryset(
-                self.filter_queryset(self.get_queryset()))
-            Notification.objects\
-                .filter(id__in=[q.id for q in queryset])\
-                .update(status=Notification.STATUS_SEEN)
-
-        return result
+    @list_route(permission_classes=[permissions.IsAuthenticated],
```
bibekdahal, line 35: You can also specify serializer_class in this decorator and user `self.get_serializer(data=request.data, many=True)` below. Specifying serializer_class here would make sure that a proper API documentation is auto-generated with correct schema.

thenav56 (approved):

PR #142, opened 2018-12-05
Change datastructure for Date Range and Time Range widget

tnagorra:
Depends on https://github.com/the-deep/client/pull/371


entry/widgets/date_range_widget.py
```diff
@@ -5,8 +5,9 @@
 
 
 def update_attribute(widget, data, widget_data):
-    from_value = data.get('from_value')
-    to_value = data.get('to_value')
+    value = data.get('value') or {}
```
frozenhelium: Is it not better to use following pattern

`value = data.get('value', {})`

PR #143, opened 2018-12-05
Add threshold for tabular autodetect and update geo detect

bewakes:


tabular/tasks.py
```diff
@@ -37,50 +42,90 @@ def _tabular_meta_extract_book(book):
...
+    # Index by codes, we need to search by codes as well
+    geos_codes = {v['code'].lower(): v for k, v in geos_names.items()}
 
     for sheet in book.sheet_set.all():
-        data = sheet.data
+        data = sheet.data or []
 
-        field_ids = data[0].keys() if data else []
-        field_ids = [x for x in field_ids if x.isnumeric()]  # omit non numeric
+        field_ids = [x for x in data[0].keys() if x.isnumeric()]
```
thenav56: if data = [], then data[0] will throw error.

bewakes: "noted with thanks". Fixed.

PR #144, opened 2018-12-06
Catch TypeError during float conversion

tnagorra:


bewakes:
This is fixed in another ongoing branch `feature-tabular-autodetect-enhancement'.

PR #171, opened 2018-12-28
Fix lead task

bewakes:


lead/receivers.py
```diff
@@ -0,0 +1,15 @@
...
+from django.dispatch import receiver
+
+from lead.models import Lead
+
+
+@receiver(models.signals.post_save, sender=Lead)
+def on_lead_saved(sender, **kwargs):
+    instance = kwargs.get('instance')
+    project = None
+    if instance:
```
thenav56: `instance` and `project` are never None.

bewakes: done

lead/tasks.py
```diff
@@ -104,6 +89,10 @@ def _extract_from_lead_core(lead_id):
             classified_doc_id=classified_doc_id,
```
thenav56: `classified_doc_id` isn't required here now.

lead/tasks.py
```diff
@@ -117,6 +106,46 @@ def _extract_from_lead_core(lead_id):
     return True
 
 
+@shared_task(bind=True, max_retries=10)
+def send_lead_text_to_deepl(self, lead_id, text):
```
thenav56: we can use `lead_id` to get `text`.

lead/tasks.py
```diff
@@ -117,6 +106,46 @@ def _extract_from_lead_core(lead_id):
...
+        }
+        response = requests.post(DEEPL_CLASSIFY_URL,
+                                 data=data)
+        response_data = response.json()
+        classified_doc_id = response_data.get('id')
+
+        # Get preview
+        preview = LeadPreview.objects.filter(lead=lead).first()
+
+        if not preview:
```
thenav56: This shouldn't happen. We should probably fail the task instead of just warn.

lead/tasks.py
```diff
@@ -117,6 +106,46 @@ def _extract_from_lead_core(lead_id):
...
+            logger.warn(
+                "Lead(id:{}) preview hasn't been created until deepl request".
+                format(lead_id)
+            )
+            return True
+
+        preview.classified_doc_id = classified_doc_id
+        preview.save()
+        return True
+    except Exception as e:
```
thenav56, line 139: This should be probably around the request only.
```python
         response = requests.post(DEEPL_CLASSIFY_URL,
                                  data=data)
         response_data = response.json()
```

lead/tasks.py
bewakes: done sir done.

lead/tasks.py
bewakes: sir yes sir.

lead/tasks.py
bewakes: right sir right

lead/tasks.py
bewakes: considered.

PR #207, opened 2019-01-31
Feature field update api

bewakes:


tabular/serializers.py
```diff
@@ -52,6 +52,7 @@ class Meta:
         exclude = ('sheet',)
 
     def get_geodata(self, obj):
+        print(obj)
```
thenav56: No print please

PR #225, opened 2019-02-20
Feature django upgrade 2.1

thenav56:
TODO: Upgrade Channels


apps/project/views.py
```diff
@@ -14,7 +14,7 @@
     views,
     viewsets,
 )
-from rest_framework.decorators import detail_route, list_route
+from rest_framework.decorators import action, action
```
bewakes: only one `action` is enough I think sir.

bewakes (changes requested):

apps/project/views.py
thenav56: Fixed sir

bewakes (approved):

PR #279, opened 2019-04-17
Add SVG for Tabular Geo and Fixes

thenav56:
* Add svg for map
* Fix recaptcha not showing in client
* Add thousands seperators [the-deep/deeper/issues/343]
* Better error handling
* Admin Panel Additions
    * Add Geo Cache Tigger Action
    * Add Search to Geo Models

### DIFF
--- a/utils/common.py
+++ b/utils/common.py
@@ -16,7 +16,6 @@ import string
 import tempfile
 import requests
 import logging
-import traceback
 
 USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1)' + \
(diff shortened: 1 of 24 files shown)


utils/common.py
```diff
@@ -279,37 +286,46 @@ def create_plotly_image(func):
     """
     def func_wrapper(*args, **kwargs):
         width, height = kwargs.pop('chart_size', (5, 4))
-        image_format = kwargs.pop('format', 'png')
+        if isinstance(kwargs.get('format', 'png'), list):
+            images_format = kwargs.pop('format', 'png')
```
bewakes, line 290: don't need to have default value `png`. can just use kwargs['format'] as well.

thenav56: @bewakes Can't use just `kwargs['format']`. Need to pop the field.
Removed the default value.

PR #288, opened 2019-05-01
Feature partial tabular

AdityaKhatri:


apps/export/views.py
```diff
@@ -70,22 +75,18 @@ def post(self, request, version=None):
                 return response.Response({}, status=status.HTTP_403_FORBIDDEN)
 
         export = Export.objects.create(
-            title='tmp',
+            title='Generating Export.....',
```
bewakes, line 78: I like the title. (y)

PR #309, opened 2019-05-30
Replace None with empty value in ary export

bewakes:


apps/ary/export/data_collection_techniques_info.py
```diff
@@ -5,6 +5,8 @@
 def format_value(val):
     if isinstance(val, list):
         return ','.join(val)
+    if val is None:
+        val = ''
```
tnagorra, line 9: very nice

PR #319, opened 2019-06-17
Add `is_private` field in the `project` model

frozenhelium:
Also exclude private projects from API results if the user does not have
access to the project


apps/project/models.py
```diff
@@ -254,7 +262,7 @@ def get_modifiable_for(user):
         ).distinct()
 
     def can_get(self, user):
-        return True
+        return Project.is_member(user)
```
bewakes: Shouldn't the condition be `is_member` or `is not private`?

apps/project/models.py
```diff
@@ -222,9 +227,12 @@ def get_annotated():
...
     @staticmethod
     def get_for(user, annotated=False):
+        # Note: `.exclude(Q(is_private=True) & ~Q(members=user)).all()`
+        # excludes the the private projects that the user is not member of
+
         if annotated:
-            return Project.get_annotated().all()
-        return Project.objects.all()
+            return Project.get_annotated().exclude(Q(is_private=True) & ~Q(members=user)).all()
+        return Project.objects.exclude(Q(is_private=True) & ~Q(members=user)).all()
```
bewakes: `.all()` is not needed here.

PR #333, opened 2019-07-03
Feature entry viz

thenav56:
TODO:
- [x] Write test.


apps/entry/stats.py
```diff
@@ -0,0 +1,264 @@
...
+            'sectors_keys': sectors_keys,
+        }
+
+
+def _get_attribute_data(collector, attribute, cd_widget_map):
+    widget_type = attribute.widget.widget_id
+    widget_pk = attribute.widget.pk
+    data = attribute.data
+
+    if widget_type not in SUPPORTED_WIDGETS or data is None or data.get('value') is [None, {}, []]:
```
bewakes: Maybe it should be `is in [None, {}, []]` instead of `is [None, {}, []]` ?

apps/project/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+def generate_entry_stats(project_id):
+    ges_wait_lock_key = GES_WAIT_LOCK_KEY.format(project_id)
+    if not redis.get_lock(ges_wait_lock_key, ProjectEntryStats.THRESHOLD_SECONDS).acquire(blocking=False):
+        # logger.warning(f'GENERATE_ENTRY_STATS:: Waiting for timeout {ges_wait_lock_key}')
+        return False
+
+    _generate_entry_stats(project_id)
+
+    # Lock for another THRESHOLD_SECONDS
+    redis.get_lock(GES_WAIT_LOCK_KEY.format(project_id), ProjectEntryStats.THRESHOLD_SECONDS).acquire(blocking=False)
```
bewakes: You can just use `ges_wait_lock_key` instead of `GES_WAIT_LOCK_KEY.format(project_id)`.
And I could not get why are we locking even after the work is done?

thenav56: Done
For allowing only one calculation per ProjectEntryStats.THRESHOLD_SECONDS interval.

apps/entry/stats.py
thenav56: Done

bewakes (approved):

PR #364, opened 2019-08-04
Fix not null contraint during clone framework

AdityaKhatri:


bewakes:
@AdityaKhatri , this has already been fixed by setting the field nullable.

AdityaKhatri:
> @AdityaKhatri , this has already been fixed by setting the field nullable.

@bewakes no it hasn't.
.get method returned null as a value to the description field and the error was thrown by database, hence an additonal check was required.

bewakes:
@AdityaKhatri migration to fix this is already in develop. Please try restarting the server.

PR #385, opened 2019-09-02
Update internal connector api and change pdna

bewakes:


bewakes:
The changes are now reflected in `feature/emm-connector` branch.

PR #389, opened 2019-09-11
Feature/lead lazy load

thenav56:


apps/lead/tests/test_apis.py
```diff
@@ -138,10 +173,72 @@ def test_update_assignee(self):
...
+
+        response = self.client.post(url, data)
         self.assert_200(response)
 
+        # Only members are returned when requested is None
+        data = {
+            'projects': [project.pk],
+        }
+        response = self.client.post(url, data)
+        assert response.json(), {
```
bewakes: This looks weird. I think it should have been `response.json() == {....`.

apps/lead/tests/test_apis.py
```diff
@@ -138,10 +173,72 @@ def test_update_assignee(self):
...
+
+        # If value are provided respective data are provided (filtered by permission)
+        data = {
+            'projects': [project.pk],
+            'leadGroups': [lead_group2.pk],
+            'members': [user1.pk, user2.pk, self.user.pk],
+            'organizations': [unhcr.pk]
+        }
+        response = self.client.post(url, data)
+        assert response.json(), {
```
bewakes: Here as well.

apps/lead/views.py
```diff
@@ -67,6 +74,19 @@ def get_queryset(self):
...
+class ProjectLeadGroupViewSet(LeadGroupViewSet):
+    """
+    NOTE: Only to be used by Project's action route [DONOT USE DIRECTLY]
+    """
+    pagination_class = AutocompleteSetPagination
+    serializer_class = SimpleLeadGroupSerializer
+    filter_backends = (filters.SearchFilter,)
+
+    def get_queryset(self):
+        project = Project.objects.get(pk=self.request.query_params['project'])
```
bewakes, line 86: Maybe use `query_params.get()`?

apps/project/tests/test_apis.py
```diff
@@ -1087,3 +1091,17 @@ def test_project_entries_stats(self):
...
+        lead_group1 = self.create(LeadGroup, project=project)
+        lead_group2 = self.create(LeadGroup, project=project)
+
+        url = f'/api/v1/projects/{project.pk}/lead-groups/'
+
+        self.authenticate()
+        response = self.client.get(url)
+        self.assert_200(response)
+        # Only provide projects leads-group [Pagination is done for larger dataset]
+        assert set([lg['id'] for lg in response.json()['results']]),\
```
bewakes: Here also. I  think it should be `assert val1 == val2`, not `assert val1, val2`.

bewakes (changes requested):

apps/lead/views.py
thenav56: It's defined behavior. It's in the `NOTE:`.
https://github.com/the-deep/server/blob/75d903f58e4947bcb1658a4295781e9f028c11b8/apps/lead/views.py#L77-L87

apps/lead/tests/test_apis.py
thenav56: Fixed

apps/lead/tests/test_apis.py
thenav56: Fixed

apps/project/tests/test_apis.py
thenav56: Fixed

PR #403, opened 2019-10-03
Fix lead options assignee

bewakes:


apps/lead/views.py
```diff
@@ -274,8 +272,8 @@ def get(self, request, version=None):
             {
                 'key': user.id,
                 'value': user.profile.get_display_name(),
-            } for user in User.objects.prefetch_related('profile').filter(
-                project_usergroup_filter
+            } for user in _filter_users_by_projects_memberships(
```
thenav56, line 275: Adding prefetch_related will make sense.

apps/lead/views.py
```diff
@@ -385,9 +376,7 @@ def post(self, request, version=None):
             # Dynamic Options
 
             'lead_groups': LeadGroup.objects.filter(project_filter, id__in=lead_groups_id).distinct(),
-            'members': (
-                User.objects.filter(id__in=members_id) if len(members_id) else User.objects
-            ).filter(project_usergroup_filter).prefetch_related('profile').distinct(),
+            'members': _filter_users_by_projects_memberships(members_qs, projects).distinct(),
```
thenav56: Adding prefetch_related will make sense.

apps/lead/views.py
bewakes: done

apps/lead/views.py
bewakes: done

thenav56 (approved):

PR #410, opened 2019-11-06
Feature/connector status

AdityaKhatri:


thenav56:
@bewakes Let's use Django auth permission generation logic for connector's source generation.
https://github.com/django/django/blob/master/django/contrib/auth/apps.py#L18-L21
https://github.com/django/django/blob/master/django/contrib/auth/management/__init__.py

bewakes:
- Create new model ConnectorSource, corresponding to connectors in source/store.
- Auto creation of ConnectorSource objects during app initialization.
- Modify Connector model's source from charfield to foreign key to Connector Source.
- Add extra key 'status' to Connector apis to state status of Sources(working or broken).

apps/connector/apps.py
```diff
@@ -1,5 +1,21 @@
 from django.apps import AppConfig
+from django.db.utils import ProgrammingError
 
 
 class ConnectorConfig(AppConfig):
     name = 'connector'
+
+    def ready(self):
+        from connector.models import ConnectorSource
```
thenav56, line 9: Use `post_migrate` signal
https://github.com/django/django/blob/master/django/contrib/auth/apps.py#L18-L21

PR #429, opened 2019-12-24
Add Polygon support in server

thenav56:
* Add polygon column for each region in Entry Export

Update

DONE:
- Assessment Export
- Add Test


apps/entry/widgets/geo_widget.py
```diff
@@ -1,5 +1,41 @@
...
+    ]
+
+
 def update_attribute(widget, data, widget_data):
-    values = data.get('value', [])
+    values = []
+    polygons = []
+
+    for geo_value in data.get('value') or []:
+        print(f'\t -> {geo_value}')
```
bewakes: print sir.

apps/entry/widgets/geo_widget.py
```diff
@@ -1,5 +1,41 @@
...
+        elif (
+            isinstance(geo_value, dict) and
+            geo_value.get('region') and
+            geo_value.get('geo_json') and
+            geo_value['geo_json'].get('properties') and
+            geo_value['geo_json']['properties'].get('title')
+        ):
+            region_id = geo_value['region']
+            title = geo_value['geo_json']['properties'].get('title')
+            print(f'\t\t -> R: {region_id} T: {title}')
```
bewakes: missing print removal :D

bewakes (changes requested):

PR #433, opened 2020-01-10
Add Entry Grouping API

thenav56:
* Add Entry Label (Project Scope) API
* Add Entry Group (Lead Scope) API


apps/entry/serializers.py
```diff
@@ -92,13 +99,64 @@ class Meta:
         fields = ('id', 'exportable', 'data',)
 
 
+class ProjectEntryLabelSerializer(DynamicFieldsMixin, UserResourceSerializer):
+    entry_count = serializers.IntegerField(read_only=True)
```
bewakes: Maybe use another serializer? Because entry_count seems to depend on queryset attribute in viewset.

thenav56: Done

PR #437, opened 2020-01-14
Add Polygon Filter for Entries

thenav56:
Usages: `/api/v1/entries/?geo_custom_shape=Line,Polygon`

### DIFF
--- a/apps/entry/filter_set.py
+++ b/apps/entry/filter_set.py
@@ -1,4 +1,3 @@
-from functools import reduce
 from datetime import datetime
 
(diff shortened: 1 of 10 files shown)


apps/entry/filter_set.py
```diff
@@ -1,3 +1,4 @@
+from functools import reduce
```
bewakes: is `reduce` being used?

thenav56: Removed

bewakes (approved):

PR #458, opened 2020-04-15
Fix/reliefweb connector

bewakes:


apps/connector/sources/relief_web.py
```diff
@@ -73,12 +279,13 @@ def fetch(self, params, offset, limit):
             post_params['limit'] = limit
 
         post_params['sort'] = ['date.original:desc', 'title:asc']
+        print(post_params)
```
AdityaKhatri: @bewakes plz remove

PR #476, opened 2020-06-30
Server Changes

thenav56:
TODO:
- [x] Ary: Clone Ary to other projects the-deep/deeper#898
- [x] Entry comments: Make assignee in entry comments multiselect the-deep/deeper#894
- [x] Organigram Widget: Show parents of selected nodes in new columns the-deep/deeper#897
  - NOTE: Will require manual migration for old organigram data. cc: @AdityaKhatri


apps/lead/views.py
```diff
@@ -651,57 +761,6 @@ def _get_clone_ready(obj, lead):
             _get_clone_ready(emm_trigger, lead) for emm_trigger in emm_triggers
         ])
 
+        if return_existing_lead:
+            return lead, True
         return lead
```
bewakes: There are two different return types. Might cause issues at the calling end.

bewakes (commented):
All is good, except for one place where a function has two different return types(one tuple and the other object).

apps/lead/views.py
thenav56: @bewakes Thanks for the review. Could you check out this commit https://github.com/the-deep/server/pull/476/commits/7d0aca86e85076f59ae6bfdcc31e9f12ae72e9dd, if it resolves?

bewakes (approved):

apps/lead/views.py
bewakes: Perfect!

PR #478, opened 2020-07-15
Add assessmentId field to leads GET API

frozenhelium:
Related to: https://github.com/the-deep/deeper/issues/908

### DIFF
--- a/apps/lead/filter_set.py
+++ b/apps/lead/filter_set.py
@@ -22,11 +22,15 @@ class LeadFilterSet(django_filters.FilterSet):
     'in' lookup expressions and CSVWidget.
     """
 
-    ENTRIES_EXISTS = 'entries_exists'
+    ENTRIES_EXIST = 'entries_exists'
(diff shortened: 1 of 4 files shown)


thenav56 (approved):

apps/lead/filter_set.py
```diff
@@ -24,9 +24,13 @@ class LeadFilterSet(django_filters.FilterSet):
 
     ENTRIES_EXISTS = 'entries_exists'
     ASSESSMENT_EXISTS = 'assessment_exists'
+    ENTRIES_DO_NOT_EXIST = 'entries_do_not_exist'
+    ASSESSMENT_DO_NOT_EXIST = 'assessment_do_not_exist'
```
bewakes: Just being a grammar Nazi here :P

bewakes (approved):

apps/lead/filter_set.py
AdityaKhatri: Yes, makes sense.
Making this consistent with entries

PR #479, opened 2020-07-15
Add entry filter by lead status

frozenhelium:
Can filter entry by lead status now using `lead_status` field

Related to: https://github.com/the-deep/deeper/issues/902

### DIFF
--- a/apps/entry/filter_set.py
+++ b/apps/entry/filter_set.py
@@ -185,7 +185,7 @@ def get_filtered_entries(user, queries):
 
     lead_status = queries.get('lead_status')
     if lead_status:
-        entries = entries.filter(lead__status=lead_status)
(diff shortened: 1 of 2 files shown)


apps/entry/filter_set.py
```diff
@@ -179,9 +179,14 @@ class Meta:
 def get_filtered_entries(user, queries):
     entries = Entry.get_for(user)
     project = queries.get('project')
+
     if project:
         entries = entries.filter(lead__project__id=project)
 
+    lead_status = queries.get('lead_status')
+    if lead_status and len(lead_status) > 0:
```
bewakes: `len(lead_status)` is redundant. `if lead_status` will suffice.

bewakes (approved):

bewakes (approved):

PR #480, opened 2020-07-15
Add organizations field

frozenhelium:
Related to: https://github.com/the-deep/deeper/issues/909

@subinasr @AdityaKhatri the 'organizations' field was changed to 'project_organization_types' in project options


apps/project/migrations/0063_auto_20200715_1131.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('project', '0062_auto_20200715_1015'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='project',
+            name='organizations',
+            field=models.ManyToManyField(blank=True, through='project.ProjectOrganization', to='organization.Organization'),
+        ),
```
bewakes, line 17: Please create a single migration for these two.
Steps:
- revert back to migration 0061 [ before org field was added] `python manage.py migrate project 0061`
- remove the newly created migration files 0062 and 0063.
- makemigration and migrate again.

**NOTE**: if you have switched branches since these migrations and ran migrations on other branches, there might be some migration issues.

bewakes (changes requested):

apps/project/migrations/0063_auto_20200715_1131.py
thenav56: cc: @AdityaKhatri @tnagorra @subinasr

apps/project/tests/test_apis.py
```diff
@@ -55,6 +61,9 @@ def test_create_project(self):
         data = {
             'title': 'Test project',
             'data': {'testKey': 'testValue'},
+            'organizations': [
+                {'organization': Organization.objects.all()[0].id, 'type': ProjectOrganization.DONOR},
```
thenav56: Mostly test is atomic in nature. So everything you create will be revert back after a test is done.
Try to create object on the same scope if possible. Basically move `self.org` inside `test_create_project` and use it directly instead of `Organization.objects.all()[0]`. This also avoids confusion on how there is an organization. Thanks.

apps/project/views.py
```diff
@@ -621,7 +622,14 @@ def get(self, request, version=None):
         if fields_query:
             fields = fields_query.split(',')
 
-        options = {}
+        options = {
+            'organizations': [
```
thenav56: Shouldn't this be like this? Just to avoid confusion. Thanks.
```suggestion
            'project_organization_types': [
```

AdityaKhatri:
@frozenhelium , could you look into the comments here?

tnagorra (changes requested):
We should include tests for this diff logic

apps/project/serializers.py
```diff
@@ -326,6 +360,33 @@ def update(self, instance, validated_data):
...
+        project_organizations = ProjectOrganization.objects.filter(project=instance)
+
+        if organizations:
+            ids = [x.id for x in project_organizations]
+            request_ids = [x.get('id') for x in organizations]
+            to_remove = list(set(ids) - set(request_ids))
+
+            for i in to_remove:
+                ProjectOrganization.objects.get(id=i).delete()
+
```
annshress: Some api tests for `update` scenario with different combinations of existing project-organizations and new project-organizations are missing.

apps/project/tests/test_apis.py
```diff
@@ -56,6 +62,9 @@ def test_create_project(self):
         data = {
             'title': 'Test project',
             'data': {'testKey': 'testValue'},
+            'organizations': [
+                {'organization': self.org1.id, 'type': ProjectOrganization.DONOR},
+            ],
         }
```
annshress, line 68: Like for create, tests for update with `project-organizations`

apps/project/serializers.py
frozenhelium: Added

thenav56:
@frozenhelium We don't need to handle nested create/update/delete action ourselves. There is an already created mixin to handle those actions.

Since the project serializer is using [UserResourceSerializer](https://github.com/the-deep/server/blob/develop/apps/user_resource/serializers.py#L10) which includes those mixin, we should be good without those custom logics.

NOTE:  Nested mixins https://github.com/the-deep/server/blob/develop/deep/writable_nested_serializers.py

frozenhelium:
@thenav56 `ProjectSerializer` already extends `UserResourceSerializer` but the organization create / update didn't work on its own. So, I did those manually.

annshress (approved):

thenav56:
> @thenav56 `ProjectSerializer` already extends `UserResourceSerializer` but the organization create / update didn't work on its own. So, I did those manually.

@frozenhelium That's weird. Have you tried without `read_only=True`?
https://github.com/the-deep/server/pull/480/files#diff-76a4e0560629a434956fbdfc05c84a72R297

frozenhelium:
@thenav56 doing that gives following error:
```
server_1  | [2020/09/28 14:20:13] HTTP PUT /api/v1/projects/1/ 500 [0.12, 172.18.0.1:47540]
server_1  | builtins.AssertionError
server_1  | Traceback (most recent call last):
server_1  |   File "/usr/local/lib/python3.6/dist-packages/rest_framework/views.py", line 492, in dispatch
server_1  |     response = handler(request, *args, **kwargs)
server_1  |   File "/usr/local/lib/python3.6/dist-packages/rest_framework/mixins.py", line 70, in update
server_1  |     self.perform_update(serializer)
server_1  |   File "/usr/local/lib/python3.6/dist-packages/rest_framework/mixins.py", line 80, in perform_update
server_1  |     serializer.save()
server_1  |   File "/code/deep/writable_nested_serializers.py", line 307, in save
server_1  |     return super().save(**kwargs)
server_1  |   File "/usr/local/lib/python3.6/dist-packages/rest_framework/serializers.py", line 209, in save
server_1  |     self.instance = self.update(self.instance, validated_data)
server_1  |   File "/code/apps/project/serializers.py", line 395, in update
server_1  |     return super().update(instance, validated_data)
server_1  |   File "/code/apps/user_resource/serializers.py", line 44, in update
server_1  |     resource = super().update(instance, validated_data)
server_1  |   File "/code/deep/writable_nested_serializers.py", line 375, in update
server_1  |     self.update_or_create_reverse_relations(instance, reverse_relations)
server_1  |   File "/code/deep/writable_nested_serializers.py", line 275, in update_or_create_reverse_relations
server_1  |     related_instance = serializer.save(**save_kwargs)
server_1  |   File "/usr/local/lib/python3.6/dist-packages/rest_framework/serializers.py", line 209, in save
server_1  |     self.instance = self.update(self.instance, validated_data)
server_1  |   File "/usr/local/lib/python3.6/dist-packages/rest_framework/serializers.py", line 973, in update
server_1  |     raise_errors_on_nested_writes('update', self, validated_data)
server_1  |   File "/usr/local/lib/python3.6/dist-packages/rest_framework/serializers.py", line 823, in raise_errors_on_nested_writes
server_1  |     class_name=serializer.__class__.__name__
server_1  | AssertionError: The `.update()` method does not support writable nested fields by default.
server_1  | Write an explicit `.update()` method for serializer `project.serializers.ProjectOrganizationSerializer`, or set `read_only=True` on nested serializer fields.
server_1  | Internal Server Error: /api/v1/projects/1/
```

frozenhelium:
@thenav56 Figured out the issue
`ProjectOrganizationSerializer` should also extend `UserResourceSerializer` for it to automatically work

annshress (approved):

PR #481, opened 2020-07-22
Fix entry comment show behavior

AdityaKhatri:


apps/templates/entry/comment_notification_email.html
```diff
@@ -97,7 +97,7 @@
     {% endif %}
     <a
         class="comment-link"
-        href="{{client_domain}}/projects/{{comment.entry.project_id}}/leads/{{comment.entry.lead_id}}/edit-entries/?entry_id={{comment.entry_id}}"
+        href="{{client_domain}}/projects/{{comment.entry.project_id}}/leads/{{comment.entry.lead_id}}/edit-entries/?entry_id={{comment.entry_id}}&show_comment=true"
```
thenav56: @AdityaKhatri Shouldn't this be in this format?

```suggestion
        href="{{client_domain}}/projects/{{comment.entry.project_id}}/leads/{{comment.entry.lead_id}}/entries/edit/?entry_id={{comment.entry_id}}&show_comment=true"
```

PR #486, opened 2020-08-13
Fix unhcr updated params

bewakes:


apps/connector/sources/unhcr_portal.py
```diff
@@ -248,13 +256,22 @@ class UNHCRPortal(Source):
...
+        resp = requests.get(url, params=params)
         return resp.text
 
     def fetch(self, params, offset, limit):
         results = []
-        if params.get('country'):
-            params['country_json'] = '{"0":"' + params['country'] + '"}'
+        country = params.pop('country', None)
+        date_from = _format_date_or_none(params.pop('date_from', None))
+        date_to = _format_date_or_none(params.pop('date_from', None))
```
annshress: Maybe its supposed to be `params.pop('date_to', None)`

bewakes: yes yes, very much !!

bewakes: Fixed.

annshress (approved):

PR #488, opened 2020-08-25
Feature/document deduplication

bewakes:
This change requires some new `ENVIRONMENT` variables to be set:
- `SOURCE_TABLE_NAME=deep-util-services-local-sources`
- `DEDUPLICATION_ES_ENDPOINT=https://search-deep-dev-7yt6m5ulk7irae7vgbgu3ueafe.us-east-1.es.amazonaws.com`
- `DEDUPLICATION_ES_STAGE=local`

The `SOURCE_TABLE_NAME` is the same value as used in the Extraction lambda function.  

@thenav56 , here is the description, let me know if anything needs to be explained in detail or needs to be changed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,7 +74,7 @@ WORKDIR /code
 COPY ./requirements.txt /code/requirements.txt
 RUN echo 'alias python=python3\nalias pip=pip3' >> ~/.bashrc \
     && python3 -m pip install --upgrade pip \ 
-    && pip3 install uwsgi \
+    && pip3 install uwsgi setuptools \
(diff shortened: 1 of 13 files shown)


bewakes:
@thenav56, there are some new environment variables being used in this feature, how should they be managed? Where should they be placed?

thenav56:
> @thenav56, there are some new environment variables being used in this feature, how should they be managed? Where should they be placed?

@bewakes Could you list the new environment variables in the PR's description. We will need to handle that here https://github.com/the-deep/deeper/

PR #493, opened 2020-09-07
Add Entry.verified status

annshress:
Related to [#932](https://github.com/the-deep/deeper/issues/932)

To verify or unverify an entry
`POST at /api/v1/entries/__EntryId__/verify/ to verify`
`POST at /api/v1/entries/__EntryId__/unverify/ to unverify`

To filter by verification status
`GET` at `/api/v1/entries/` with request parameters `?verified=True` or `?verified=False`


AdityaKhatri (approved):

AdityaKhatri:
@annshress , 
1. let's add the field `verification_last_changed_by` to save the user who verified it or unverified it.
2. Let's also add the behavior that whenever a user edits the entry that is verified, we automatically change it to an unverified state.
3. Also, let's change the API to:
/verify and /unverify instead of query params.

thenav56 (approved):

PR #504, opened 2020-09-25
Add entries filter for lead confidentiality and lead priority

frozenhelium:
- [x] Lead priority
- [x] Lead confidentiality
- [x] Lead published on

### DIFF
--- a/apps/entry/filter_set.py
+++ b/apps/entry/filter_set.py
@@ -20,7 +20,7 @@ from entry.models import (
 
 # We don't use UserResourceFilterSet since created_at and modified_at
 # are overridden below
-class EntryFilterSet(django_filters.FilterSet):
+class EntryFilterSet(django_filters.rest_framework.FilterSet):
     """
     Entry filter set
     Basic filtering with lead, excerpt, lead title and dates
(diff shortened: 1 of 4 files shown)


apps/entry/filter_set.py
```diff
@@ -187,6 +209,28 @@ def get_filtered_entries(user, queries):
...
+    lead_confidentiality = queries.get('lead_confidentiality')
+    if lead_confidentiality:
+        entries = entries.filter(lead__confidentiality__in=lead_confidentiality)
+
+    lead_published_on_lt = queries.get('lead_published_on__lt')
+    print('>>>>>>>>>>>>>>>>>>>>>>>>>', lead_published_on_lt)
+    # if lead_published_on_lt:
+    #     entries = entries.filter(
+    #         lead__published_on__lt=datetime.date(lead_published_on_lt),
+    #     )
```
tnagorra: Maybe we can remove these?

apps/entry/filter_set.py
```diff
@@ -187,6 +209,28 @@ def get_filtered_entries(user, queries):
...
+
+    lead_published_on_gte = queries.get('lead_published_on__gte')
+    print('>>>>>>>>>>>>>>>>>>>>>>>>>', lead_published_on_gte)
+    # if lead_published_on_gte:
+    #     entries = entries.filter(
+    #         lead__published_on__gte=datetime.date(lead_published_on_gte),
+    #     )
```
tnagorra: Maybe we can remove these?

frozenhelium: not yet

apps/entry/filter_set.py
frozenhelium: not yet

thenav56 (approved):

PR #518, opened 2020-10-06
Feature/bulk lead delete

annshress:
Bulk deleting leads

Endpoints:
```js
POST 
/api/v1/project/<PROJECT-ID>/leads/dry-bulk-delete/ 
data: {leads: <LIST of lead ids>}
response: number of entities that will cascade delete (for now: entries, assessments)

POST
/api/v1/project/<PROJECT-ID>/leads/bulk-delete/ 
data: {leads: <LIST of lead ids>}
response: 204 No Content
```

### DIFF
--- a/deep/permissions.py
+++ b/deep/permissions.py
@@ -68,41 +68,25 @@ class CreateLeadPermission(permissions.BasePermission):
 class DeleteLeadPermission(permissions.BasePermission):
     """Checks if user can delete lead(s)"""
     def has_permission(self, request, view):
-        if request.method not in ('DELETE',):
+        if request.method not in ('POST', 'DELETE'):
             return True
 
-        project_id = request.data.get('project')
-        lead_ids = request.data.get('ids', '').split(',')
+        project_id = view.kwargs.get('project_id')
 
         if not project_id:
-            return False
-        if not lead_ids:
-            return True
-
-        # Since this supports list value for project
-        if isinstance(project_id, list):
-            project_ids = set(project_id)
-        else:
-            project_ids = {project_id}
-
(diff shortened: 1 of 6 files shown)


AdityaKhatri:
@annshress , lets use array instead of comma-separated ids, also let's rename ids to leads.
Let's scope the URL by project Id as well.

thenav56:
> @annshress , lets use array instead of comma-separated ids, also let's rename ids to leads.
> Let's scope the URL by project Id as well.

Yeah, scoping by project will make the permission check easier too.

deep/permissions.py
```diff
@@ -65,6 +65,46 @@ def has_permission(self, request, view):
         return projects_count == len(project_ids)
 
 
+class DeleteLeadPermission(permissions.BasePermission):
+    """Checks if user can delete lead(s)"""
+    def has_permission(self, request, view):
+        if request.method not in ('DELETE',):
+            return True
+
+        project_id = request.parser_context['kwargs']['project']
```
annshress: @bewakes  Can you verify if this permission handler is okay?

frozenhelium (approved):

deep/permissions.py
bewakes: this seems okay. But just to be on the safe side, use `.get('project')`.

bewakes: However, I do not understand(or maybe if I forgot if it was there already) why we are passing project id. Shouldn't leads be sufficient to get project information?

tnagorra: I think we are limiting a user's action to a specific project first.
We shouldn't be able to bulk delete leads of different projects. @thenav56 right?

PR #521, opened 2020-10-07
Add authors in project viz api

AdityaKhatri:


thenav56 (approved):

apps/entry/stats.py
```diff
@@ -27,6 +27,14 @@ def _get_lead_data(lead):
             'id': lead.source.id,
             'title': lead.source.data.title,
         },
+        'author_raw': lead.author_raw,
```
annshress, line 30: Isn't this legacy? deprecation prone or not? @thenav56

annshress (approved):

apps/entry/stats.py
annshress: > purano ho, tara hatauna parne type ko haina kyare
- @AdityaKhatri

thenav56: This is legacy but we can't remove it until we migrate all raw author string to organization reference. Since it a lot of work we will fix this later when required (which looks to be far in the future). We can use `author_raw` search when we do this later instead of `TODO` for each author_raw uses.

PR #539, opened 2020-11-03
Feature/unified connector

bewakes:


apps/connector/models.py
```diff
@@ -165,3 +168,150 @@ def save(self, *args, **kwargs):
...
+                    )
+                ).order_by().values('published_date').annotate(
+                    count=models.Count('*'),
+                ).values('count', date=models.F('published_date'))
+            ),
+        }
+        if commit:
+            self.save()
+
+    def save(self, *args, **kwargs):
```
bewakes, line 300: Maybe need to update `self.old_params` as `self.old_params = self.params`. Because if not, the next `.save()` call on the same object will result in inconsistencies.

apps/connector/sources/store.py
```diff
@@ -14,17 +14,24 @@
...
+    emm.EMM,
+    acaps_briefing_notes.AcapsBriefingNotes,
+    unhcr_portal.UNHCRPortal,
+    pdna.PDNA,
+    research_center.ResearchResourceCenter,
+    wpf.WorldFoodProgramme,
+    humanitarian_response.HumanitarianResponse,
+]
+
+assert len(set([source.key for source in sources_list])) == len(sources_list), 'Duplicate key found for connectors'
```
bewakes, line 30: I think it's too strict to assert here although this get caught very fast. Whole app won't work(and no test is run) if here's some issue. Maybe shift this to test cases? Given that we only deploy after all the testcases pass, it's both less strict and caught fast enough when done in test.

bewakes (commented):
@thenav56  please look at the comments. I could not request changes because I opened the PR, haha.

apps/connector/sources/store.py
thenav56: This assertion is for hardcoded values to avoid duplicate keys for each connector source. 

The reason to be strict is that it needs to be strict :stuck_out_tongue:.

apps/connector/models.py
thenav56: Nice catch. Also found the same in lead save.

annshress (approved):

annshress (approved):

PR #563, opened 2020-11-09
Fix/lead attributes plural

samshara:


thenav56:
@samshara Please update the test too.
Also, These changes should be reflected in clients too. cc: @AdityaKhatri @frozenhelium @tnagorra.

samshara:
@annshress please fix the test too

PR #566, opened 2020-11-10
Feature/filtered entries summary

annshress:


apps/entry/views.py
```diff
@@ -49,7 +51,31 @@
...
+            total_unique_authors=total_unique_org_type,
+            total_sources=total_sources,
+        )
+        return response.Response(OrderedDict([
+            ('count', self.paginator.count),
+            ('next', self.paginator.get_next_link()),
+            ('previous', self.paginator.get_previous_link()),
+            ('results', data),
+            ('summary', summary_data)
+        ]))
```
thenav56: Can we do it like this?
```suggestion
        return response.Response({
            **super().get_paginated_response(data).data,   # or just data
            'summary': summary_data,
        })
```

thenav56 (approved):

apps/entry/views.py
```diff
@@ -49,7 +52,41 @@
...
 
 
-class EntryViewSet(viewsets.ModelViewSet):
+class EntrySummaryPaginationMixin(object):
+    def get_paginated_response(self, data):
+        if not self.request.data.get('calculate_summary',
+                                     self.request.GET.get('calculate_summary', '0')) == '1':
+            return super().get_paginated_response(data)
+        qs = self.filter_queryset(self.get_queryset())
+        q = qs.annotate(org=models.functions.Coalesce('lead__authors__parent', 'lead__authors')).values('org').annotate(org_type=models.functions.Coalesce('lead__authors__parent__organization_type', 'lead__authors__organization_type')).values('org_type').order_by('org_type').annotate(count=models.Count('org', distinct=True)).values('org_type', 'count')
```
thenav56: Very long. Please break this to multiple lines.
```suggestion
        q = qs.annotate(
                org=models.functions.Coalesce('lead__authors__parent', 'lead__authors')
            ).values('org').annotate(
                org_type=models.functions.Coalesce(
                    'lead__authors__parent__organization_type',
                    'lead__authors__organization_type'
                )
            ).values('org_type').order_by('org_type').annotate(
                count=models.Count('org', distinct=True)
            ).values('org_type', 'count')
```

apps/entry/views.py
```diff
@@ -49,7 +52,41 @@
...
+            if each.get('count') and each.get('org_type'):
+                org_type = OrganizationType.objects.get(id=each['org_type'])
+                org_type_count.append(
+                    dict(org=dict(
+                       id=each.get('org_type'),
+                       title=org_type.title,
+                       shortName=org_type.short_name,
+                    ),
+                    count=each.get('count'))
+                )
```
thenav56: I am not sure but you should be able to get this data directly from the queryset.
```suggestion
        org_type_count = list(
            q.values('count', id=F('org_type__id'), title=F('org_type__title'), short_name=F('ort_type__short_name'))
        )
```

Make sure to remove the last `.values(` from q declaration. Also, if it throws an error like **there is a column already**, then use list comprehension for defining dict manually.

thenav56: Also, try not to use camelCase in python. Specifed DRF renderer will do that for us (if `list(` is not used then DRF renderer may not convert the snake_case to CamelCase)

thenav56 (approved):

PR #568, opened 2020-11-10
Feature/widget exports

annshress:
export API will expect, `exporting_widgets={List of widget identifiers}` within existing `filters` to export widget data.

Run the following to update exportdata of appropriate widgets.

```
python manage.py update_attribute_for_widget --widget scaleWidget
```
dateWidget or timeWidget or dateRangeWidget


apps/entry/widgets/conditional_widget.py
```diff
@@ -52,9 +53,9 @@ def update_attribute(widget, data, widget_data):
 
         if w_export_data:
             excel_data.append(w_export_data.get('data', {}).get('excel'))
-            report_keys += w_export_data.get('data', {})\
-                .get('report', {})\
-                .get('keys') or []
+            report_data = w_export_data.get('data', {})\
+                .get('report', {})
+            report_keys = report_data.get('keys') or []
```
thenav56: There is `+=` before?

annshress: I found it was not necessary. `report_keys = []` was directly used just there.

apps/export/entries/report_exporter.py
```diff
@@ -173,18 +210,58 @@ def _add_widget_information_into_report(self, para, report):
...
+            site=settings.DEEPER_FRONTEND_HOST,
+            project=entry.lead.project.id,
+            lead=entry.lead.id,
+            entry=entry.id
+        )
+        # format of exporting_widgets = "[517,43,42,[405,"scalewidget-xe11vlcxs2gqdh1r","Scale"]]"
+        widget_keys = [Widget.objects.get(id=each).key \
+            if not isinstance(each, list) else each[1] \
+            for each in self.exporting_widgets
+        ]
```
thenav56:
```suggestion
        widget_keys = [
            Widget.objects.get(id=each).key if not isinstance(each, list) else each[1]
            for each in self.exporting_widgets
        ]
```

apps/export/entries/report_exporter.py
```diff
@@ -173,18 +210,58 @@ def _add_widget_information_into_report(self, para, report):
...
+        # para.add_hyperlink(url, f"{entry.lead.id}-{entry.id}")
+        para.add_run(f'{entry.lead.id}-{entry.id}', bold=True)
+        export_data = []
+        for each in entry.exportdata_set.all():
+            if 'other' in each.data.get('report', {}):
+                for rep in each.data['report'].get('other', {}):
+                    if rep.get('widget_key') and rep['widget_key'] in widget_keys:
+                        export_data.append(rep)
```
thenav56: `rep` will be a key?

annshress: report.other is a list.. Nice catch

apps/entry/widgets/date_widget.py
```diff
@@ -22,13 +23,19 @@ def update_attribute(widget, data, widget_data):
 
         'export_data': {
             'data': {
-                'excel': {
+                'version': 1,
+                'common': {
                     'value': date,
+                    'widget_id': WIDGET_ID,
+                    'widget_key': widget.key
+                },
```
annshress: @thenav56 I have defined meta data here. because during attribute export, it solely relies on `export_data.data`. Not wanting to produce some unwanted side-effects

apps/entry/widgets/date_range_widget.py
```diff
@@ -35,9 +36,16 @@ def update_attribute(widget, data, widget_data):
 
         'export_data': {
             'data': {
-                'excel': {
+                'version': 1,
```
thenav56: The version also needs to be like **WIDGET_ID**, since we will use it to list which attributes need updating.

Also, we should add this as a column or at the root of **export_data**.

apps/entry/widgets/date_range_widget.py
```diff
@@ -35,9 +36,16 @@ def update_attribute(widget, data, widget_data):
...
+                'common': {
                     'values': [from_date, to_date],
+                    'widget_id': WIDGET_ID,
+                    'widget_key': widget.key,
+                },
+                'excel': {
                 },
+                'report': {
+                }
```
thenav56, line 52: If we are using `.get('excel', {})` in export, then we will not need this. What do you think?

annshress: Thought about that. But left it as so such that next viewer will not have to go through the code to know the schema of the json data.

PR #572, opened 2020-11-25
Add AND filter in frameworks list widgets

AdityaKhatri:


annshress (approved):

apps/entry/filter_set.py
```diff
@@ -287,15 +290,25 @@ def get_filtered_entries(user, queries):
...
+            query = query_and or query
             if not isinstance(query, list):
                 query = query.split(',')
 
-            if len(query) > 0:
-                entries = entries.filter(
-                    filterdata__filter=filter,
-                    filterdata__values__overlap=query,
-                )
+            elif len(query) > 0:
```
tnagorra: Shouldn't this be `if` instead of `elif`

tnagorra (approved):

PR #588, opened 2020-12-08
Add verified entries count in lead filtered viewset

annshress:


apps/lead/models.py
```diff
@@ -250,6 +250,10 @@ def get_associated_entities(cls, project_id, lead_ids):
             'assessments': Assessment.objects.filter(lead__project_id=project_id, lead__in=lead_ids).count(),
         }
 
+    @property
+    def verified_entries_count(self):
+        return self.entry_set.filter(verified=True).count()
+
```
thenav56: This is okay but we should use annotate to calculate the count on the listing view.

apps/lead/views.py
```diff
@@ -207,6 +207,7 @@ def leads_filter(self, request, version=None):
         filter_data = self._get_processed_filter_data(raw_filter_data)
 
         qs = LeadFilterSet(data=filter_data, queryset=self.get_queryset()).qs
+        qs.prefetch_related('entry_set')
```
thenav56: Just for the count, this will not be necessary (also high memory usage). Use annotate instead.

thenav56 (approved):

PR #589, opened 2020-12-08
Filter entry list by lead title v2

annshress:


apps/entry/filter_set.py
```diff
@@ -143,6 +148,13 @@ class Meta:
             },
         }
 
+    # def lead_title_filter(self, queryset, name, value):
+    #     if value.strip():
+    #         return queryset.annotate(
+    #             idx=StrIndex(Lower('lead__title'), Value(value.lower()))
+    #         ).filter(idx__gt=0).order_by('idx', 'lead__title')
+    #     return queryset
+
```
thenav56, line 157: Let's not keep commented code. If you think it's good to keep this, let's create a 2 commit w/o this.

apps/entry/filter_set.py
```diff
@@ -120,13 +121,17 @@ class EntryFilterSet(django_filters.rest_framework.FilterSet):
         label='Lead Group Label',
         method='lead_group_label_filter',
     )
+    lead__title = django_filters.CharFilter(
+        field_name='lead__title',
+        lookup_expr='icontains'
+    )
```
thenav56, line 127: Let's add a test for this.

apps/entry/filter_set.py
tnagorra: I second @thenav56

AdityaKhatri:
@annshress , lets close this.

PR #591, opened 2020-12-08
Leads: Filter by authoring organization type

k9845:


apps/lead/filter_set.py
```diff
@@ -5,8 +5,10 @@
 from project.models import Project
 from user.models import User
 from .models import Lead, LeadGroup
-
-
+from organization.models import (
+    Organization,
+    OrganizationType
+)
```
thenav56: Let's import in order like this. (newline in-between)

- Python native modules
- External modules
- Local modules
-Relative modules

So move this before `from .models...`

apps/lead/filter_set.py
```diff
@@ -126,6 +132,7 @@ class Meta:
             'emm_risk_factors': ['exact'],
             'created_at': ['exact', 'lt', 'gt', 'lte', 'gte'],
             'published_on': ['exact', 'lt', 'gt', 'lte', 'gte'],
+            'authoring_organization': ['exact'],
```
annshress: Lets remove this. Redundant.

annshress (approved):

apps/lead/views.py
```diff
@@ -375,13 +381,15 @@ def post(self, request, version=None):
         serializer = LeadOptionsBodySerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
         fields = serializer.data
+        print(fields)
```
AdityaKhatri: remove this

apps/lead/filter_set.py
```diff
@@ -114,6 +114,10 @@ class LeadFilterSet(django_filters.FilterSet):
         method='ordering_filter',
     )
 
+    authoring_organization = django_filters.CharFilter(
+        method='organization_type_filter'
+    )
+
```
thenav56: Let's do this the same way we are doing for the entries. Also `authoring_organizations`

apps/lead/views.py
```diff
@@ -382,6 +388,7 @@ def post(self, request, version=None):
         emm_entities = fields['emm_entities']
         emm_keywords = fields['emm_keywords']
         emm_risk_factors = fields['emm_risk_factors']
+        organization_type_id = fields['organization_types']
```
thenav56: This should be `organization_type_ids`

thenav56 (approved):

apps/lead/tests/test_apis.py
```diff
@@ -982,6 +985,52 @@ def test_lead_filter(self):
...
+        url = '/api/v1/leads/?authoring_organization_types={}'
+
+        project = self.create_project(title="lead_test_project")
+        organization_type1 = self.create(OrganizationType, title="National")
+        organization_type2 = self.create(OrganizationType, title="International")
+        organization_type3 = self.create(OrganizationType, title="Government")
+
+        organization1 = self.create(Organization, organization_type=organization_type1)
+        organization2 = self.create(Organization, organization_type=organization_type2)
+        organization3 = self.create(Organization, organization_type=organization_type3)
```
thenav56, line 998: Let's add another organization with parent.

apps/lead/models.py
```diff
@@ -223,6 +223,10 @@ def get_for(cls, user):
             assessment_id=models.F('assessment'),
             verified_entries_count=models.Count('entry',
                                                 filter=models.Q(entry__verified=True)),
+            organization_types=models.functions.Coalesce(
+                'authors__parent__organization_type',
+                'authors__organization_type'
+            ),
```
thenav56: Let's move this to the custom filter function. Wouldn't make sense to calculate this when not required.

thenav56 (approved):

PR #606, opened 2020-12-14
Users: Restrict users API access to unauthenticated users

k9845:


apps/user/permissions.py
```diff
@@ -0,0 +1,16 @@
...
+class UserPermission(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        if request.method in permissions.SAFE_METHODS:
+            return True
+        return obj == request.user
+
+
+class UserViewSetPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if (request.method == 'POST' or (request.user and request.user.is_authenticated)):
```
annshress: Use `view.action == 'create'` rather than `request.method`.

apps/user/views.py
```diff
@@ -25,13 +25,7 @@
...
     PasswordResetSerializer,
 )
-
-
-class UserPermission(permissions.BasePermission):
-    def has_object_permission(self, request, view, obj):
-        if request.method in permissions.SAFE_METHODS:
-            return True
-        return obj == request.user
+from .permissions import UserPermission, UserViewSetPermission
```
annshress, line 28: Nice!

annshress (approved):

PR #613, opened 2020-12-17
Put password character limit to 128

k9845:


apps/user/tests/test_apis.py
```diff
@@ -178,3 +181,15 @@ def test_user_preference_feature_access(self):
...
+class PasswordCheckerTest(TestCase):
+
+    def test_password_greater_than_128_characters(self):
+        expected_error = "This password has exceed the limit of %d characters"
+        self.assertIsNone(CustomMaximumLengthValidator().validate('12345678'))
+        self.assertIsNone(CustomMaximumLengthValidator(max_length=20).validate('123'))
+
+        with self.assertRaises(ValidationError) as vd:
+            CustomMaximumLengthValidator(max_length=128).validate('12'*129)
+        self.assertEqual(vd.exception.messages, [expected_error % 128])
```
annshress, line 195: lets not keep it in `api tests`

deep/settings.py
```diff
@@ -198,6 +198,9 @@
         'NAME': 'django.contrib.auth.password_validation.'
         'NumericPasswordValidator',
     },
+    {
+        'NAME': 'user.validators.CustomMaximumLengthValidator',
+    },
```
thenav56, line 203: `AUTH_PASSWORD_VALIDATORS` is only used on Django password create/change level https://docs.djangoproject.com/en/3.1/topics/auth/passwords/#module-django.contrib.auth.password_validation

We also need to check this when a user is created through API and when a user tries to login.

PR #614, opened 2020-12-18
Export UI: Filter exports by status and archived state

k9845:


annshress (approved):

apps/export/tests/test_apis.py
```diff
@@ -56,3 +56,17 @@ def test_trigger_api_with_export_permission(self):
         export = Export.objects.get(id=response.data['export_triggered'])
         self.assertTrue(export.pending)
         self.assertEqual(export.exported_by, self.user)
+
+    def test_export_filter_by_status(self):
```
thenav56, line 60: We will also need a test where a client archives an export.
Also, @AdityaKhatri Can a user unarchive an export?

PR #622, opened 2020-12-22
Allow entry filters with lead filters in lead/filters api

annshress:
Add two more filter params to existing `lead filter api`

- entries_filter: json = this expects the same filters as Entry list filter
- **custom_filters**: choices [ `exclude_empty_filtered_entries`, `exclude_empty_verified_filtered_entries` ]
    - if `exclude_empty_filtered_entries` will not return leads with zero filtered entries
    - if `exclude_empty_verified_filtered_entries` will not return leads with zero filtered AND verified entries

filtered entries count can be retrieved from `filteredEntriesCount` and similarly `verifiedFilteredEntriesCount`

### DIFF
--- a/apps/lead/filter_set.py
+++ b/apps/lead/filter_set.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.core.exceptions import FieldError
 import django_filters
 
 from deep.filter_set import DjangoFilterCSVWidget
@@ -33,6 +34,10 @@ class LeadFilterSet(django_filters.FilterSet):
         (ASSESSMENT_EXISTS, 'Assessment Exists'),
         (ENTRIES_DO_NOT_EXIST, 'Entries do not exist'),
         (ASSESSMENT_DOES_NOT_EXIST, 'Assessment does not exist'),
+    )
+    EXCLUDE_EMPTY_FILTERED_ENTRIES = 'exclude_empty_filtered_entries'
+    FILTERED_ENTRIES_CHOICE = (
+        (EXCLUDE_EMPTY_FILTERED_ENTRIES, 'exclude empty filtered entries'),
     )
 
     search = django_filters.CharFilter(method='search_filter')
@@ -120,6 +125,11 @@ class LeadFilterSet(django_filters.FilterSet):
         method='authoring_organization_types_filter',
         widget=DjangoFilterCSVWidget,
         queryset=OrganizationType.objects.all(),
+    )
+    # used in export
+    custom_filters = django_filters.ChoiceFilter(
+        label='Filtered Exists Choice',
+        choices=FILTERED_ENTRIES_CHOICE, method='filtered_exists_filter',
     )
 
     class Meta:
@@ -220,6 +230,13 @@ class LeadFilterSet(django_filters.FilterSet):
                 )
             )
             return qs.filter(organization_types__in=[ot.id for ot in value]).distinct()
+        return qs
+
+    def filtered_exists_filter(self, qs, name, value):
+        if value == self.EXCLUDE_EMPTY_FILTERED_ENTRIES:
+            # NOTE: `filtered_entries_count` is annotated in LeadViewSet.filter, currently used for
+            # solely export list page
+            return qs.filter(filtered_entries_count__gte=1)
         return qs
 
 
--- a/apps/lead/filter_set.py
+++ b/apps/lead/filter_set.py
@@ -36,8 +36,10 @@ class LeadFilterSet(django_filters.FilterSet):
         (ASSESSMENT_DOES_NOT_EXIST, 'Assessment does not exist'),
     )
     EXCLUDE_EMPTY_FILTERED_ENTRIES = 'exclude_empty_filtered_entries'
+    EXCLUDE_EMPTY_VERIFIED_FILTERED_ENTRIES = 'exclude_empty_verified_filtered_entries'
     FILTERED_ENTRIES_CHOICE = (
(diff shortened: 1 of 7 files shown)


apps/lead/tests/test_apis.py
```diff
@@ -1066,6 +1067,32 @@ def test_lead_filter_search(self):
...
+        assert response.json()['count'] == 0, 'There are not supposed to be leads with entries'
+
+        self.create(Entry, project=project, lead=lead1, verified=True)
+        self.create(Entry, project=project, lead=lead1, verified=False)
+
+        post_data = {'filtered_entries_exists': 'exclude_empty_filtered_entries',
+                     'entries_filter': {'verified': True}}
+        response = self.client.post(url, post_data)
+        assert response.json()['count'] == 1
+        print(response.json())
```
thenav56:
```suggestion
```

apps/lead/filter_set.py
```diff
@@ -34,6 +35,12 @@ class LeadFilterSet(django_filters.FilterSet):
         (ENTRIES_DO_NOT_EXIST, 'Entries do not exist'),
         (ASSESSMENT_DOES_NOT_EXIST, 'Assessment does not exist'),
     )
+    INCLUDE_EMPTY_FILTERED_ENTRIES = 'include_empty_filtered_entries'
+    EXCLUDE_EMPTY_FILTERED_ENTRIES = 'exclude_empty_filtered_entries'
+    FILTERED_ENTRIES_CHOICE = (
+        (INCLUDE_EMPTY_FILTERED_ENTRIES, 'include empty filtered entries'),
```
thenav56: Exclude should be enough for now. We will add other custom filters later.
```suggestion
```

apps/lead/filter_set.py
```diff
@@ -121,6 +128,11 @@ class LeadFilterSet(django_filters.FilterSet):
         widget=DjangoFilterCSVWidget,
         queryset=OrganizationType.objects.all(),
     )
+    # used in export
+    filtered_entries_exists = django_filters.ChoiceFilter(
+        label='Filtered Exists Choice',
+        choices=FILTERED_ENTRIES_CHOICE, method='filtered_exists_filter',
+    )
```
thenav56: `custom_filters` or `complex_filters`?

apps/lead/filter_set.py
```diff
@@ -222,6 +234,15 @@ def authoring_organization_types_filter(self, qs, name, value):
             return qs.filter(organization_types__in=[ot.id for ot in value]).distinct()
         return qs
 
+    def filtered_exists_filter(self, qs, name, value):
+        if value == self.EXCLUDE_EMPTY_FILTERED_ENTRIES:
+            try:
+                # annotated in LeadViewSet
+                return qs.filter(filtered_entries_count__gte=1)
+            except FieldError:
+                return qs
```
thenav56: We have to make sure this is always provided. Let's remove try/except and make sure through test case.

apps/lead/models.py
```diff
@@ -260,6 +260,10 @@ def get_verified_entries_count(self):
             return self.verified_entries_count
         return self.entry_set.filter(verified=True).count()
 
+    def get_filtered_entries_count(self):
+        if hasattr(self, 'filtered_entries_count'):
+            return self.filtered_entries_count
+        return self.entry_set.count()
```
thenav56: This function is returning different values as entry_set doesn't have `filtered entries`.

k9845 (approved):

apps/lead/models.py
```diff
@@ -218,6 +223,29 @@ def get_for(cls, user):
...
+                models.Subquery(
+                    get_filtered_entries(user, entries_filter_data).filter(
+                        lead=models.OuterRef('pk'),
+                        verified=True
+                    ).annotate(
+                        count=models.Count('id')
+                    ).values('count')[:1], output_field=models.IntegerField()
+                ), 0
+            ),
+        )
```
thenav56: If `entries_filter_data` is empty, Can we use the `no_of_entries` which is calculated below?
Add this annotate after the current annotation.

Also, let's change `no_of_entries` to `entries_count`. cc: @AdityaKhatri 

Something like this?
```suggestion
        entries_filter_data = filters.get('entries_filter_data', {})
        return qs.annotate(
            no_of_entries=models.Count('entry', distinct=True),
            assessment_id=models.F('assessment'),
            verified_entries_count=models.Count('entry', filter=models.Q(entry__verified=True)),

            filtered_entries_count=(
                models.functions.Coalesce(
                    models.Subquery(
                        get_filtered_entries(user, entries_filter_data).filter(
                            lead=models.OuterRef('pk')
                        ).annotate(
                            count=models.Count('id')
                        ).values('count')[:1], output_field=models.IntegerField()
                    ), 0
                ) if entries_filter_data else F('no_of_entries')
            ),
            filtered_verified_entries_count=(
                models.functions.Coalesce(
                    models.Subquery(
                        get_filtered_entries(user, entries_filter_data).filter(
                            lead=models.OuterRef('pk'),
                            verified=True
                        ).annotate(
                            count=models.Count('id')
                        ).values('count')[:1], output_field=models.IntegerField()
                    ), 0
                ) if entries_filter_data else F('verified_entries_count')
            )
        )
```

AdityaKhatri: Okay, lets make it consistent

thenav56 (approved):

thenav56 (approved):

annshress:
Closed.

Combined with #629 .

PR #625, opened 2020-12-23
Export UI: Add api to delete/archive an export

k9845:


apps/export/tests/test_apis.py
```diff
@@ -56,3 +56,17 @@ def test_trigger_api_with_export_permission(self):
...
+    def test_delete_export(self):
+        export = self.create(Export, exported_by=self.user)
+        url = '/api/v1/exports/{}/'.format(export.id)
+
+        self.authenticate()
+        response = self.client.delete(url)
+        self.assert_204(response)  # delete from api
+
+        # check for database
+        assert Export.objects.count() == 1  # should not delete from database
```
thenav56: Instead of `1` calculate the count before DELETE request and then make sure they are the same.

apps/export/views.py
```diff
@@ -26,14 +27,23 @@ def get(self, request, file_id=None, version=None):
...
+class ExportViewSet(mixins.DestroyModelMixin,
+                    mixins.ListModelMixin,
+                    mixins.RetrieveModelMixin,
+                    viewsets.GenericViewSet):
     serializer_class = ExportSerializer
     permission_classes = [permissions.IsAuthenticated]
     filterset_class = ExportFilterSet
 
     def get_queryset(self):
         return Export.get_for(self.request.user)
```
thenav56, line 35: This should filter out the soft-deleted Exports. Add filter is_deleted=False in get_for function's queryest.

thenav56: Also, add test for this too.

apps/export/views.py
```diff
@@ -26,14 +27,23 @@ def get(self, request, file_id=None, version=None):
         pass
 
 
-class ExportViewSet(viewsets.ReadOnlyModelViewSet):
+class ExportViewSet(mixins.DestroyModelMixin,
+                    mixins.ListModelMixin,
+                    mixins.RetrieveModelMixin,
+                    viewsets.GenericViewSet):
```
thenav56: We can now use `ModelViewSet` here directly. Make sure the permission is defined correctly.

annshress: But it will also expose, `create` and `update` mixins. Maybe this is okay? @thenav56 @k9845

thenav56: Yeah. But we need `update` for `is_archived`.
And for `create` we could disable it for now or move the `ExportTriggerView` to `ExportViewSet::create`.

annshress (approved):

AdityaKhatri (approved):

apps/export/views.py
```diff
@@ -6,6 +6,7 @@
     views,
     viewsets,
     status,
+    mixins
```
thenav56: Unused import.

apps/export/permissions.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework import permissions
+
+
+class ExportViewSetPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if (view.action in ('list', 'update', 'destroy') or (request.user and request.user.is_authenticated)):
+            return True
+        return False
```
thenav56: After updating `get_for` we would not need any additional permission other than `is_authenticated`. So we should remove this. @annshress let me know if I am missing something. @k9845 by this, use previous permission_classes in the viewset.

Also `(view.action in ('list', 'update', 'destroy') or (request.user and request.user.is_authenticated))` can you explain why is this here?

apps/export/tests/test_apis.py
```diff
@@ -56,3 +56,19 @@ def test_trigger_api_with_export_permission(self):
         export = Export.objects.get(id=response.data['export_triggered'])
         self.assertTrue(export.pending)
         self.assertEqual(export.exported_by, self.user)
+
```
thenav56, line 59: We also need additional tests to make sure deleted exports are not available in list/retrieve/update APIs too. And non-owner users can't access other's exports.

apps/export/models.py
```diff
@@ -75,12 +75,14 @@ class Export(models.Model):
...
 
     def __str__(self):
         return self.title
 
     @staticmethod
     def get_for(user):
         return Export.objects.filter(
-            exported_by=user
+            exported_by=user,
+            is_deleted=False
```
thenav56, line 87: @AdityaKhatri Do you think we also need to filter access by project permission? If a user is not a member of a project they don't have access to their exports from that project?

annshress (approved):

annshress (approved):

thenav56 (approved):

PR #629, opened 2020-12-28
Allow entry filters with lead filters in lead/filters api; Filter leads and inclusion parameter during export

annshress:
Add two more filter params to existing `lead filter api`

- entries_filter: json = this expects the same filters as Entry list filter
- custom_filters: choices [ `exclude_empty_filtered_entries`, `exclude_empty_verified_filtered_entries` ]
    - if `exclude_empty_filtered_entries` will not return leads with zero filtered entries
    - if `exclude_empty_verified_filtered_entries` will not return leads with zero filtered AND verified entries

filtered entries count can be retrieved from `filteredEntriesCount` and similarly `verifiedFilteredEntriesCount`.

...

Previously exports only required `lead = [...ids]` param. Now it requires following new params along with existing `lead`.

1. `include_leads`: Boolean = it states whether to include or exclude the `lead` from export.
2. `lead_filters`= Same as lead filter api, along with a nested `entries_filters` key to filter the entries.

~~Depends on~~ #622 rebased and closed.

### DIFF
--- a/apps/lead/views.py
+++ b/apps/lead/views.py
@@ -149,7 +149,9 @@ class LeadViewSet(viewsets.ModelViewSet):
 
     def get_queryset(self):
         filters = dict()
-        filters['entries_filter_data'] = self.request.data.pop('entries_filter', {})
+        filters['entries_filter_data'] = {
+            f[0]: f[1] for f in self.request.data.pop('entries_filter', [])
+        }
         leads = Lead.get_for(self.request.user, filters)
         lead_id = self.request.GET.get('similar')
         if lead_id:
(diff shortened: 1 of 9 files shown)


k9845 (approved):

apps/lead/views.py
```diff
@@ -8,6 +8,7 @@
 from django.contrib.postgres.search import TrigramSimilarity
 from django.db import models, transaction
 
+from deep import compiler
```
thenav56:
```suggestion
from deep import compiler  # noqa: F401
```

https://flake8.pycqa.org/en/3.7.9/user/error-codes.html

thenav56 (approved):

PR #641, opened 2021-01-11
Fix ProjectMembership Api

k9845:
1.Only project members can see project memberships of that project
    2.Only project admins can add users of that project
    3.Only project admins can modify users of that project

### DIFF
--- a/apps/project/permissions.py
+++ b/apps/project/permissions.py
@@ -109,7 +109,6 @@ class MembershipModifyPermission(permissions.BasePermission):
             member=request.user
         ).first()
         user_role = user and user.role
-
         if not user_role or user_role.level > obj.role.level:
             return False
 
--- a/apps/project/serializers.py
+++ b/apps/project/serializers.py
@@ -325,6 +325,14 @@ class ProjectSerializer(RemoveNullFieldsMixin,
         model = Project
         exclude = ('members', )
 
+    def to_representation(self, data):
+        result = super().to_representation(data)
+        if not result['is_private'] and self.context['request'].user.id in [e['member'] for e in result['memberships']]:
+            pass
+        else:
+            result['memberships'] = []
+        return result
+
     def create(self, validated_data):
         member = self.context['request'].user
         is_private = validated_data.get('is_private', False)
@@ -457,6 +465,8 @@ class ProjectSerializer(RemoveNullFieldsMixin,
                 'Invalid analysis framework: {}'.format(analysis_framework.id))
         return analysis_framework
 
+    """def to_representation(self, data):
+        print("data", data)"""
 
 class ProjectStatSerializer(ProjectSerializer):
     number_of_leads = serializers.IntegerField(read_only=True)
--- a/apps/project/serializers.py
+++ b/apps/project/serializers.py
@@ -286,11 +286,6 @@ class ProjectUsergroupMembershipSerializer(RemoveNullFieldsMixin,
 
 class ProjectSerializer(RemoveNullFieldsMixin,
                         DynamicFieldsMixin, UserResourceSerializer):
-    memberships = ProjectMembershipSerializer(
-        source='projectmembership_set',
-        many=True,
-        read_only=True,
-    )
 
     organizations = ProjectOrganizationSerializer(
         source='projectorganization_set',
@@ -324,14 +319,6 @@ class ProjectSerializer(RemoveNullFieldsMixin,
     class Meta:
         model = Project
         exclude = ('members', )
-
-    def to_representation(self, data):
-        result = super().to_representation(data)
-        if not result['is_private'] and self.context['request'].user.id in [e['member'] for e in result['memberships']]:
-            pass
-        else:
-            result['memberships'] = []
-        return result
 
     def create(self, validated_data):
         member = self.context['request'].user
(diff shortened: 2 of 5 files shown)


apps/project/serializers.py
```diff
@@ -457,6 +465,8 @@ def validate_analysis_framework(self, analysis_framework):
                 'Invalid analysis framework: {}'.format(analysis_framework.id))
         return analysis_framework
 
+    """def to_representation(self, data):
+        print("data", data)"""
```
thenav56:
```suggestion
```

apps/project/tests/test_apis.py
```diff
@@ -700,9 +719,51 @@ def test_add_member_duplicate(self):
 
         self.authenticate()
         response = self.client.post(url, data)
+        print(response.data)
```
thenav56:
```suggestion
```

apps/project/tests/test_apis.py
```diff
@@ -38,7 +38,7 @@
 
 
 class ProjectApiTest(TestCase):
-    fixtures = ['apps/ary/fixtures/ary_template_data.json']
+    fixtures = ['ary_template_data.json']
```
thenav56, line 41: Could you explain why this is needed?

apps/project/views.py
```diff
@@ -480,9 +480,12 @@ def get_members(self, request, pk=None, version=None):
...
         ).distinct()
-        self.page = self.paginate_queryset(members)
-        serializer = self.get_serializer(self.page, many=True)
-        return self.get_paginated_response(serializer.data)
+        if members.filter(id=request.user.id).exists():
+            self.page = self.paginate_queryset(members)
+            serializer = self.get_serializer(self.page, many=True)
+            return self.get_paginated_response(serializer.data)
+        else:
+            return response.Response('Forbidden Acess', status=403)
```
thenav56: You can use `IsProjectMember` in **permission_classes** instead of custom logic here.

apps/project/serializers.py
```diff
@@ -325,6 +325,14 @@ class Meta:
...
         exclude = ('members', )
 
+    def to_representation(self, data):
+        result = super().to_representation(data)
+        if not result['is_private'] and self.context['request'].user.id in [e['member'] for e in result['memberships']]:
+            pass
+        else:
+            result['memberships'] = []
+        return result
+
```
thenav56: Instead of this, we should create 2 serializers (one for members and one for public view) and only show members for members. This way we can add additional fields for member view later also. You can create a **ProjectSerailizer** for non-member and inheritance class **ProjectMemberViewSerializer** (or better naming?) for members with additional fields.
cc: @annshress 

@AdityaKhatri Is member data used in the client from project API? If not we can remove this entirely for now.

apps/project/tests/test_apis.py
k9845: giving full path was causing error while running tests @thenav56 dai.

thenav56: Okay

apps/project/permissions.py
```diff
@@ -84,6 +84,15 @@ def has_object_permission(self, request, view, obj):
         )
 
 
+class IsProjectMember(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        from project.models import ProjectMembership
+        return ProjectMembership.objects.filter(
+            project=obj,
+            member=request.user
+        ).exists()
```
thenav56: This already exists.

Let's do this.
- Move `has_object_permission` logic there.
- Also use `project.members.filter` directly
- Make sure to check if `obj` is a project object.

https://github.com/the-deep/server/blob/17244762287d5b2fca2bdadd5017a5402730c602/deep/permissions.py#L156-L170

apps/project/tests/test_apis.py
```diff
@@ -74,19 +74,6 @@ def test_create_project(self):
...
-                         self.user.pk)
-
-        # assert single membership
-        self.assertEqual(len(response.data['memberships']), 1)
-        self.assertEqual(len(response.data['organizations']), 1)
-        membership = ProjectMembership.objects.get(
-            pk=response.data['memberships'][0]['id'])
-        self.assertEqual(membership.member.pk, self.user.pk)
-        self.assertEqual(membership.role, self.admin_role)
-
```
thenav56, line 89: Is this moved elsewhere?

deep/permissions.py
```diff
@@ -167,4 +167,9 @@ def has_permission(self, request, view):
             return True
         elif entry_id and Entry.get_for(request.user).filter(id=entry_id).exists():
             return True
-        return False
+        return True
```
thenav56, line 170: This will always return **True** even if the member is not part of the project.

If `project_id` `lead_id` `entry_id` is defined then we should make sure the user is part of the project. For that, change the logic above to something like this.
```python
        if project_id:
            return Project.get_for_member(request.user).filter(id=project_id).exists()
```

apps/project/views.py
```diff
@@ -136,6 +139,15 @@ class ProjectViewSet(viewsets.ModelViewSet):
     def get_queryset(self):
         return get_filtered_projects(self.request.user, self.request.GET)
 
+    def get_serializer_class(self):
+        # get the project and check for its member with current user
+        if (self.lookup_url_kwarg or self.lookup_field) not in self.kwargs:
```
annshress, line 144: `if self.action == "list"`

annshress (approved):

annshress (approved):

thenav56 (approved):

Issue #598, opened 2020-12-14
Entry Excerpt: Migrate old image data to new mechanism


AdityaKhatri:
Affected apis:
- users/me/preferences/
- users/
- users/2/
- combined/?search=H&*members_exclude_project=2&apis=users%2Cuser-groups
- notifications/

Also need to verify all attachment objects have file property.

thenav56:
Code is pushed for review. But real data migration needs to be done after it's pushed to production.

Issue #596, opened 2020-12-14
Entry Excerpt: Add field `lead_image` in API


AdityaKhatri:
Only allow id of lead preview to be sent in lead_image field
The file will be later copied to Gallery File model and it will be sent back in image field

PR #652, opened 2021-01-20
Feature/entry image fix

thenav56:
NOTE: Client using file API might break.

Command to migrate data to the new structure: `python3 manage.py bulk_migrate entry_images_v2`

Issues:
-  Entry Excerpt: Add field `lead_image` in API #596
- Lead Preview: Send lead image file ID #595 
- Entry Excerpt: Only allow ID of file instead of URL in image field #594 
- Entry Excerpt: Migrate old image data to new mechanism #598 
- Gallery file: Provide only files uploaded by user via API #597


apps/bulk_data_migration/entry_images_v2/migrate.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def _get_file_from_file_url(entry, string):
+    fileid = parse_number(string.rstrip('/').split('/')[-1])
+    return fileid and File.objects.filter(id=fileid).first()
+    try:
+        fileid = parse_number(string.rstrip('/').split('/')[-1])
+        return fileid and File.objects.filter(id=fileid).first()
+    except IndexError:
+        pass
```
annshress: Maybe, we only need `try...except` block here.

k9845 (approved):

apps/lead/models.py
```diff
@@ -226,7 +226,10 @@ def get_for(cls, user):
         )
 
     def get_assignee(self):
-        return self.assignee.first()
+        # To be used by prefetch_related (Also, there is only one assignee)
+        assignees = self.assignee.all()
+        if assignees:
+            return assignees[0]
```
annshress, line 265: `self.assignee.first()` will suffice

AdityaKhatri: Please read the note above

apps/bulk_data_migration/entry_images_v2/migrate.py
```diff
@@ -0,0 +1,85 @@
...
+def _get_file_from_s3_url(entry, string):
+    file_path = '/'.join(string.split('?')[0].split('/')[4:])
+    # NOTE: For lead-preview generate gallery files
+    if file_path.startswith('lead-preview/'):
+        lead_preview = LeadPreviewImage.objects.filter(file=file_path).first()
+        if lead_preview and lead_preview.file.storage.exists(lead_preview.file):
+            return lead_preview.clone_as_deep_file(entry.created_by)
+        return
+    try:
+        file_path = '/'.join(string.split('?')[0].split('/')[4:])
```
annshress: Redundant. This try block should have been at the top instead.

apps/bulk_data_migration/entry_images_v2/migrate.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _get_file_from_s3_url(entry, string):
+    file_path = '/'.join(string.split('?')[0].split('/')[4:])
+    # NOTE: For lead-preview generate gallery files
+    if file_path.startswith('lead-preview/'):
+        lead_preview = LeadPreviewImage.objects.filter(file=file_path).first()
+        if lead_preview and lead_preview.file.storage.exists(lead_preview.file):
+            return lead_preview.clone_as_deep_file(entry.created_by)
+        return
+    try:
```
annshress, line 40: Just like `lead-preview`, an explicit `if file_path .starts_with(...)` would be concrete. But if you think its fine, please resolve.

apps/lead/models.py
thenav56: Basically `self.assignee.first()` will not use prefetch cache.

apps/bulk_data_migration/entry_images_v2/migrate.py
thenav56: Done

apps/bulk_data_migration/entry_images_v2/migrate.py
thenav56: Done

annshress (approved):

k9845 (approved):

apps/bulk_data_migration/entry_images_v2/migrate.py
```diff
@@ -0,0 +1,82 @@
...
+    except IndexError:
+        return
+    return fileid and File.objects.filter(id=fileid).first()
+
+
+def _get_file_from_s3_url(entry, string):
+    try:
+        file_path = '/'.join(string.split('?')[0].split('/')[4:])
+    except IndexError:
+        return
```
annshress, line 34: Is the use of `...[4:]` okay? Just to be safe, if things change, and we always get the `IndexError`.

annshress (approved):

apps/bulk_data_migration/entry_images_v2/migrate.py
thenav56: According to the data currently in production, this should work without any issue. Also, there is image_raw with the original value which we can use to analyze and fix remaining if found.

PR #655, opened 2021-01-27
Add model for Assignment

k9845:
1. Assignment API  `/api/v1/assignments/`
2. Assignment  filter by project `/api/v1/assignments/?{project.id}`
3. Mark `is_done` for particular assignment `/api/v1/assignments/{assignment.id}`
4. Mark all `is_done`  `/api/v1/assignments/bulk-mark-as-done/`

@AdityaKhatri @samshara @subinasr


apps/notification/receivers/entry_comment.py
```diff
@@ -89,3 +95,45 @@ def create_entry_commit_text_notification(sender, instance, created, **kwargs):
...
+    if action == 'post_add' and pk_set:
+        for receiver in pk_set:
+            Assignment.objects.create(
+                content_object=instance,
+                created_for_id=receiver,
+                project=instance.entry.project,
+            )
+
+    if action == 'post_remove' and pk_set:
+        Assignment.objects.filter(entry__id=instance.id).delete()
```
annshress: This will DELETE all the assignments of that entry. I do not think we want that!

apps/notification/receivers/entry_comment.py
```diff
@@ -89,3 +95,45 @@ def create_entry_commit_text_notification(sender, instance, created, **kwargs):
...
+
+
+@receiver(post_delete, sender=Lead)
+def delete_assignment(sender, instance, *args, **kwargs):
+    assignment = instance.id
+    Assignment.objects.filter(lead__id=assignment).delete()
```
annshress:
```suggestion
    lead_id = instance.id
    Assignment.objects.filter(lead__id=lead_id).delete()
```

apps/notification/views.py
```diff
@@ -71,3 +71,34 @@ def get_count(self, request, version=None):
...
+        return Assignment.get_for(
+            self.request.user,
+        )
+
+    def filter_queryset(self, queryset):
+        qs = super().filter_queryset(queryset)
+        project = self.request.query_params.get('project')
+        if project is not None:
+            qs.filter(project=project)
+        return qs
```
thenav56: Let's use django-filter FilterSet.

apps/notification/receivers/entry_comment.py
```diff
@@ -89,3 +95,45 @@ def create_entry_commit_text_notification(sender, instance, created, **kwargs):
...
+
+
+@receiver(post_delete, sender=EntryComment)
+def delete_assignment(sender, instance, *args, **kwargs):
+    assignment = instance.id
+    Assignment.objects.filter(entry__id=assignment).delete()
```
annshress:
```suggestion
    entry_id = instance.id
    Assignment.objects.filter(entry__id=entry_id).delete()
```

apps/notification/tests/test_apis.py
```diff
@@ -223,3 +225,89 @@ def test_get_notification_count(self):
...
+
+    def test_get_assignments_lead(self):
+        project = self.create_project()
+        project1 = self.create_project()
+        user1 = self.create(User)
+        user2 = self.create(User)
+
+        url = '/api/v1/assignments/'
+
+        self.authenticate()
```
annshress:
```suggestion
        self.authenticate(user1)
```

apps/notification/views.py
```diff
@@ -71,3 +71,34 @@ def get_count(self, request, version=None):
...
+        permission_classes=[permissions.IsAuthenticated],
+        url_path='status'
+    )
+    def get_status(self, request, version=None):
+        queryset = self.filter_queryset(self.get_queryset()).filter(is_done=False)
+        temp = list(queryset.values_list("id", flat=True))  # lazy evaluation of queryset
+        queryset.update(is_done=True)
+        data = Assignment.objects.filter(id__in=temp)
+        serializer = AssignmentSerializer(data, many=True)
+        return response.Response(serializer.data)
```
thenav56: - Let's use POST method here.
- I think we don't have to return the assignments for bulk in this case. @AdityaKhatri @samshara Do we?
    - We can just return the number of records we changed.

apps/notification/views.py
```diff
@@ -71,3 +71,34 @@ def get_count(self, request, version=None):
...
+
+class AssignmentViewSet(viewsets.ModelViewSet):
+    serializer_class = AssignmentSerializer
+    permission_classes = [permissions.IsAuthenticated]
+    filterset_fields = ('project',)
+
+    def get_queryset(self):
+        return Assignment.get_for(
+            self.request.user,
+        )
```
thenav56: @AdityaKhatri @samshara Do we need to filter out the already done assignments or client will request with a filter?

apps/notification/models.py
```diff
@@ -73,3 +75,31 @@ class Meta:
...
         return Notification.objects.filter(receiver=user).distinct()
+
+
+class Assignment(models.Model):
+    """
+    Assignment Model
+    """
+    timestamp = models.DateTimeField(
+        default=timezone.now,
+    )
```
thenav56: Let's use auto_now_add here.
https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.DateField.auto_now_add

Also, can we change the timestamp to `created_at`? We are using that naming for most of the other models.

apps/notification/tests/test_notification.py
```diff
@@ -161,3 +162,83 @@ def test_notification_updated_on_request_aborted(self):
...
+
+        lead.assignee.remove(user1, user2)
+        lead.save()
+        self.assertEqual(Assignment.objects.count(), 0)
+
+        # try to get the assignment for user
+        lead.assignee.add(user1, user2)
+        lead.save()
+        assignment = Assignment.get_for(user1)
+        assert assignment.count() == 1  # for only the user
```
annshress, line 244: Lets add test for:
- What happens to assignment, when we delete the lead and entry comment.

apps/entry/serializers.py
```diff
@@ -418,6 +418,21 @@ class Meta:
...
+    project = serializers.IntegerField(source='entry.project_id', read_only=True)
+    project_title = serializers.CharField(source='entry.project.title', read_only=True)
+    created_by_name = serializers.CharField(source='created_by.username', read_only=True)
+    entry_excerpt = serializers.CharField(source='entry.excerpt', read_only=True)
+
+    class Meta:
+        model = EntryComment
+        fields = ('content_id', 'content_title', 'entry',
+                  'project', 'project_title', 'created_by',
+                  'created_by_name', 'entry_excerpt')
```
thenav56: Let's define this in notification/serializer.py.

Also, we don't need to define project fields here, let's use the assignment project column directly.

apps/lead/serializers.py
```diff
@@ -109,6 +109,21 @@ class Meta:
...
+    content_title = serializers.CharField(source='title', read_only=True)
+    project_title = serializers.CharField(source='project.title', read_only=True)
+    created_by_name = serializers.CharField(source='created_by.username', read_only=True)
+
+    class Meta:
+        model = Lead
+        fields = (
+            'content_id', 'content_title', 'project',
+            'project_title', 'created_by', 'created_by_name',
+        )
```
thenav56: Same as EntryCommentSerialier.

apps/notification/models.py
```diff
@@ -73,3 +75,31 @@ class Meta:
...
+    created_for = models.ForeignKey(User, on_delete=models.CASCADE)
+    project = models.ForeignKey(
+        Project,
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True,
+        default=None
+    )
```
thenav56: If others haven't pulled this change. Could you revert your migration in your system, remove old migrations files and create a fresh one where project null=True is not allowed?

apps/notification/models.py
```diff
@@ -73,3 +75,31 @@ class Meta:
...
+    is_done = models.BooleanField(default=False)
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    content_object = GenericForeignKey('content_type', 'object_id')
+
```
thenav56, line 104: let's also define the `created_by` field, where we save the user whose request triggered this assignment?
cc: @annshress @AdityaKhatri @samshara

annshress (changes requested):

apps/notification/receivers/entry_comment.py
thenav56:
```suggestion
    Assignment.objects.filter(lead__id=instance.id).delete()
```

apps/notification/receivers/entry_comment.py
```diff
@@ -89,3 +95,45 @@ def create_entry_commit_text_notification(sender, instance, created, **kwargs):
...
+    pk_set = kwargs.get('pk_set', [])
+    if action == 'post_add' and pk_set:
+        for receiver in pk_set:
+            Assignment.objects.create(
+                content_object=instance,
+                created_for_id=receiver,
+                project=instance.project,
+            )
+
+    if action == 'post_remove' and pk_set:
```
thenav56:
```suggestion
    elif action == 'post_remove' and pk_set:
```

apps/notification/tests/test_notification.py
thenav56, line 244: Let's also add a test for delete cases.

thenav56 (changes requested):

apps/notification/receivers/entry_comment.py
k9845: @annshress  I don't think so , as this action will handle the changed assignee that have been changed either from lead or entry comment, since this checks the list of assignees id that comes in `pk_set`. Also added the test case in `notification/tests/test_notification.py` go through that.

apps/notification/views.py
AdityaKhatri: Yes, a list of successful ids will be best.

apps/notification/views.py
AdityaKhatri: Best to have the client side do it, in case we need the list of already done assignments in future.

apps/notification/models.py
AdityaKhatri: Yes, that is needed.

apps/notification/views.py
thenav56: @AdityaKhatri Do we really need the list of ids? Are we storing assignments in redux? If we are re-loading after a bulk update, the count should be fine (for showing users).
Let's try to limit data when doing heavy bulk operations.

apps/notification/views.py
thenav56: @k9845 Let's add the is_done filter in the filterset too.

apps/notification/views.py
AdityaKhatri: Okay, that sounds good.
I was thinking of only updating the local state, but we can pull the data, it's fine.

Conclusion:
Only require a count of successfully updated records.

k9845: Count it is @AdityaKhatri

apps/notification/views.py
k9845: @thenav56 
done

apps/notification/models.py
k9845: Done

apps/notification/models.py
```diff
@@ -73,3 +75,37 @@ class Meta:
...
+    )
+    created_for = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        related_name='created_for',
+    )
+    project = models.ForeignKey(
+        Project,
+        on_delete=models.CASCADE,
+        blank=True,
```
thenav56: `blank=True` is not required here. Removing this will create another migration again, so revert back and create a single migration.

Ref: https://docs.djangoproject.com/en/3.1/ref/models/fields/#blank

apps/notification/receivers/entry_comment.py
```diff
@@ -1,14 +1,20 @@
...
 
 from entry.models import EntryComment, EntryCommentText
 from entry.serializers import EntryCommentSerializer
-from notification.models import Notification
+from notification.models import Notification, Assignment
 from notification.tasks import send_entry_comment_email
-
+from lead.models import Lead
 import logging
+from datetime import datetime
```
thenav56: Let's move these two to the top.

apps/notification/receivers/entry_comment.py
```diff
@@ -89,3 +95,49 @@ def create_entry_commit_text_notification(sender, instance, created, **kwargs):
...
+@receiver(m2m_changed, sender=Lead.assignee.through)
+def lead_assignment_signal(sender, instance, action, **kwargs):
+    pk_set = kwargs.get('pk_set', [])
+    if action == 'post_add' and pk_set:
+        for receiver in pk_set:
+            Assignment.objects.create(
+                content_object=instance,
+                created_for_id=receiver,
+                project=instance.project,
+                created_by=instance.created_by,
```
thenav56: We should use `request.user` instead of an instance.created_by. Take a condition where a user1 changes a lead create by user2.

For getting the request. You can either use a library module or a custom code.
- https://pypi.org/project/django-currentuser/
- https://github.com/IFRCGo/go-api/blob/develop/middlewares/middlewares.py#L12-L48

apps/notification/serializers.py
```diff
@@ -37,3 +43,58 @@ def get_data(self, notification):
             return notification.data
 
         return {}
+
+
+class AssignmentEntryCommentSerializer(serializers.ModelSerializer):
+    content_id = serializers.IntegerField(source='id', read_only=True)
+    content_title = serializers.CharField(source='text', read_only=True)
+    project = serializers.IntegerField(source='entry.project_id', read_only=True)
+    project_title = serializers.CharField(source='entry.project.title', read_only=True)
```
thenav56: we don't need project data from entry. We can use Assignment Project directly avoiding unnecessary project fetch through the entry.

thenav56 (changes requested):

annshress (approved):

apps/notification/views.py
```diff
@@ -71,3 +75,29 @@ def get_count(self, request, version=None):
...
+    serializer_class = AssignmentSerializer
+    permission_classes = [permissions.IsAuthenticated]
+    filter_backends = (django_filters.rest_framework.DjangoFilterBackend,)
+    filterset_class = AssignmentFilterSet
+    pagination_class = AssignmentPagination
+
+    def get_queryset(self):
+        return Assignment.get_for(self.request.user).select_related(
+            'project', 'created_by', 'content_type',
+        )
```
thenav56, line 90: @k9845 Could you check if select_related is working here?. Basically enable silk, add `USE_SILK=True` to .env and `docker-compose up -d server`. Then you will be able to see the request /silk.

Then hit the API endpoint and check if there is n+1 query.

k9845: Checked with silk and found no any n+1 query. @thenav56

thenav56: Super

thenav56 (approved):

annshress (approved):

PR #703, opened 2021-03-22
Project/entry review

thenav56:


apps/notification/tasks.py
```diff
@@ -24,3 +27,50 @@ def send_entry_comment_email(user_id, comment_id):
...
+            'CommentType': CommentType,
+            'notification_type': notification_type,
+            'comment': comment,
+        },
+        subject_template_name='entry/review_comment_notification_email.txt',
+        email_template_name='entry/review_comment_notification_email.html',
+    )
+
+
+def send_notifications_for_commit(comment_pk, meta):
```
k9845: `send_notifications_for_comment` might be better, what say?

apps/project/models.py
```diff
@@ -365,6 +378,8 @@ class ProjectMembership(models.Model):
         null=True, blank=True, default=None,
         related_name='added_project_memberships',
     )
+    # Represents additional permission like QA
+    badges = ArrayField(enum.EnumField(BadgeType), default=list, blank=True)
```
k9845, line 382: This must be removed in favour of a later one or is it intended?

thenav56: Both are required, `ProjectMembership` is a lower instance and `ProjectUserGroupMembership` is an instance that creates multiple `ProjetMembership` instances.

k9845: Okay, we better add a comment for these cases .

apps/notification/tasks.py
thenav56: Good Catch

apps/project/models.py
thenav56: It should be self-explanatory. But added a comment as well.

k9845 (dismissed):

PR #717, opened 2021-04-06
Add role_title and addedby_username in project memberships and usergroup api

k9845:
Changed API:
- for project-memberships from `/api/v1/project-memberships` to `/api/v1/projects/{project.id}/project-memberships/`
- for project-usergroups from `/api/v1/project-usergroups` to `/api/v1/projects/{project.id}/project-usergroups/`

CC: @samshara  @subinasr

### DIFF
--- a/apps/project/serializers.py
+++ b/apps/project/serializers.py
@@ -128,6 +128,7 @@ class ProjectMembershipSerializer(RemoveNullFieldsMixin,
         read_only=True,
         many=True,
     )
+    role_title = serializers.CharField(source='role.title', read_only=True)
 
     class Meta:
         model = ProjectMembership
@@ -446,6 +447,8 @@ class ProjectJoinRequestSerializer(RemoveNullFieldsMixin,
 
 class ProjectUserGroupSerializer(serializers.ModelSerializer):
     title = serializers.CharField(source='usergroup.title', read_only=True)
+    role_title = serializers.CharField(source='role.title', read_only=True)
+    added_by_name = serializers.CharField(source='added_by.profile.get_display_name', read_only=True)
 
     class Meta:
         model = ProjectUserGroupMembership
--- a/apps/project/serializers.py
+++ b/apps/project/serializers.py
@@ -148,17 +148,14 @@ class ProjectMembershipSerializer(RemoveNullFieldsMixin,
             raise serializers.ValidationError('Invalid project')
         return project
 
+    def project_member_validation(self, project, member):
+        if ProjectMembership.objects.filter(project=project).filter(member=member).exists():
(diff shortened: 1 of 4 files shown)


annshress (approved):

apps/project/serializers.py
```diff
@@ -147,18 +149,24 @@ def validate_project(self, project):
...
+        data['project_id'] = int(self.context['view'].kwargs['project_id'])
+        # get the members of the project
+        members = ProjectMembership.objects.filter(
+            project=data['project_id']
+        ).values('member')
+        member = data.get('member')
+        if member:
+            member = member.id
+        if member in [mem['member'] for mem in members]:
+            raise serializers.ValidationError({'member': 'Member already exist'})
```
annshress: Separate this validation to a different function.

Further, use something like below,

```
if ProjectMembership.objects.filter(project=data['project_id']).filter(member=data.get('member')).exists():
    raise ...
```

apps/project/serializers.py
```diff
@@ -128,10 +128,12 @@ class ProjectMembershipSerializer(RemoveNullFieldsMixin,
         read_only=True,
         many=True,
     )
+    role_title = serializers.CharField(source='role.title', read_only=True)
```
thenav56: Let's define SimpleRoleSerializer and use that instead.

role_details = SimpleRoleSerializer(source='role', read_only=True)

Make sure to also use select related. So role_id and role_title should be enough.

apps/project/serializers.py
```diff
@@ -128,10 +128,12 @@ class ProjectMembershipSerializer(RemoveNullFieldsMixin,
...
 
     class Meta:
         model = ProjectMembership
         fields = '__all__'
+        read_only_fields = ('project', )
```
thenav56:
```suggestion
        read_only_fields = ('project',)
```

apps/project/serializers.py
```diff
@@ -446,10 +451,17 @@ def create(self, validated_data):
 
 class ProjectUserGroupSerializer(serializers.ModelSerializer):
     title = serializers.CharField(source='usergroup.title', read_only=True)
+    role_title = serializers.CharField(source='role.title', read_only=True)
```
thenav56: Let's define SimpleRoleSerializer and use that instead.

role_details = SimpleRoleSerializer(source='role', read_only=True)

Make sure to also use select related. So role_id and role_title should be enough.

thenav56 (approved):

apps/project/serializers.py
k9845: Done!

annshress (approved):

Issue #712, opened 2021-04-01
Export: Remove empty `[]` if none of the entry metadata are selected


AdityaKhatri:
Create a new issue for the solution.
Discuss with team if we can include the solution in this sprint.

Max hours: 4 hours

Issue #711, opened 2021-04-01
Export: Add option to remove entry data


AdityaKhatri:
New options are:
1. Remove entry id
2. Remove ary details
3. Remove entries metadata

subinasr:
@Navin-Ayer I've used these names for the corresponding filters, let me know if they're okay
show_entry_id:  showEntryId
show_ary_details: showAryDetails
show_additional_metadata: showAdditionalMetadata

thenav56:
Thanks @Subina-Sharma-Rajopadhyaya 
Can you use this instead, as the filters are only for the `report` export type?

```
report_show_lead_entry_id
report_show_assessment_data
report_show_entry_widget_data
```

PR #718, opened 2021-04-07
Update API to include skip attributes for report export

thenav56:
- Export: Add option to remove entry data #711 
- Export: Remove empty `[]` if none of the entry metadata are selected #712

### DIFF
--- a/apps/export/entries/report_exporter.py
+++ b/apps/export/entries/report_exporter.py
@@ -56,18 +56,35 @@ class ExportDataVersionMismatch(Exception):
 
 
 class ReportExporter:
-    def __init__(self, exporting_widgets=None, is_preview=False):
+    def __init__(
+        self,
+        exporting_widgets=None,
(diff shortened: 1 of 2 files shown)


k9845 (approved):

annshress (approved):

apps/export/entries/report_exporter.py
```diff
@@ -55,19 +55,167 @@ class ExportDataVersionMismatch(Exception):
...
 
 
+class WidgetExporter:
+
+    @staticmethod
+    def _add_common(para, text, bold):
+        para.add_run(text, bold)
+
+    @staticmethod
+    def _add_scale_widget_data(para, label, color, bold):
```
bimal125, line 65: Make sure return is missing or its intentionally done.

thenav56: Return is not required here. This function is using para to add attributes to the document.

PR #726, opened 2021-04-14
Add project_status in project-options viewset

k9845:
fixes:
- #725 Project Options: Send options for project status in `api/v1/project-options/`

### DIFF
--- a/apps/project/tests/test_apis.py
+++ b/apps/project/tests/test_apis.py
@@ -2,6 +2,7 @@ import uuid
 
 from dateutil.relativedelta import relativedelta
 from django.utils import timezone
+from django.utils.hashable import make_hashable
 
 from user.models import (
     User,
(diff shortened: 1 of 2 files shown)


apps/project/views.py
```diff
@@ -864,7 +864,12 @@ def _filter_by_projects(qs, projects):
                 {'key': 'my_projects', 'value': 'My projects'},
                 {'key': 'not_my_projects', 'value': 'Not my projects'}
             ]
-
+        options['project_status'] = [
+            {
+                'key': status[0],
+                'value': status[1]
+            } for status in Project.STATUS_CHOICES
+        ]
```
thenav56: Let's use this.
```suggestion
        options['project_status'] = [
            {
                'key': value,
                'value': label
            } for value, label in Project.STATUS_CHOICES
        ]
```

apps/project/tests/test_apis.py
```diff
@@ -798,6 +798,18 @@ def test_options(self):
...
+        url = '/api/v1/project-options/'
+
+        self.authenticate()
+        response = self.client.get(url)
+        self.assert_200(response)
+        self.assertIn('project_status', response.data)
+        self.assertEqual(response.data['project_status'][0]['key'], 'active')
+        self.assertEqual(response.data['project_status'][0]['value'], 'Active')
+        self.assertEqual(response.data['project_status'][1]['key'], 'inactive')
+        self.assertEqual(response.data['project_status'][1]['value'], 'Inactive')
```
thenav56: Let's use variables instead of raw values.


```python3
from django.utils.hashable import make_hashable

choices = dict(make_hashable(Project.STATUS_CHOICES))

# For `active`/`inactive` let's use Project.ACTIVE Project.INACTIVE

# For `Active`/`Inactive` let's use choices[Project.ACTIVE] / choices[Project.INACTIVE]
```

k9845: Done, Thanks

thenav56 (approved):

annshress (approved):

PR #739, opened 2021-04-20
Add api for the entries list on post of entries id

k9845:
Fixes:
- #739  API to fetch list of entry data when list of entry ids is sent

### DIFF
--- a/apps/entry/filter_set.py
+++ b/apps/entry/filter_set.py
@@ -134,9 +134,10 @@ class EntryFilterSet(django_filters.rest_framework.FilterSet):
     )
 
     # Entries Id
-    entry_ids = django_filters.CharFilter(
+    entry_ids = django_filters.ModelMultipleChoiceFilter(
         label='entry_filter',
-        method='entry_ids_filter'
+        method='entry_ids_filter',
+        queryset=Entry.objects.all()
     )
 
     class Meta:
@@ -220,9 +221,8 @@ class EntryFilterSet(django_filters.rest_framework.FilterSet):
         return qs
 
     def entry_ids_filter(self, qs, name, value):
-        value = eval(value)
-        if type(value) is list:
-            return qs.filter(id__in=value)
+        if value:
+            return qs.filter(id__in=[entry.id for entry in value])
         return qs
 
     @property
--- a/apps/entry/filter_set.py
+++ b/apps/entry/filter_set.py
@@ -133,6 +133,12 @@ class EntryFilterSet(django_filters.rest_framework.FilterSet):
         queryset=OrganizationType.objects.all(),
     )
 
+    # Entries Id
+    entry_ids = django_filters.CharFilter(
+        label='entry_filter',
+        method='entry_ids_filter'
+    )
+
     class Meta:
         model = Entry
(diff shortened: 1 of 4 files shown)


annshress (approved):

apps/entry/views.py
```diff
@@ -282,6 +282,21 @@ def unverify_entry(self, request, **kwargs):
...
+        return Entry.objects.all()
+
+    def post(self, request, version=None):
+        entries_id = request.data['ids']
+        entries = Entry.objects.filter(id__in=entries_id)
+        serializer_class = self.get_serializer_class()
+        serializer = serializer_class(entries, many=True, context={'request': request})
+        return response.Response(serializer.data)
+
+
```
thenav56: No No No. As specified before let's only add a filter to the already existing entry API. So that user can filter by entry id as well.

Also, make sure users only have access to the entry they should.

Let's add a custom filter to entryfilterset, then add a test where it should filter using entry id. Also, add another test where user ask for non-authorized entry as well.

annshress: Makes more sense!

annshress (changes requested):

apps/entry/tests/test_apis.py
```diff
@@ -682,6 +682,57 @@ def test_authoring_organization_filter(self):
...
+            'filters': [[k, v] for k, v in filters.items()]
+        }
+
+        self.authenticate(user)
+        response = self.client.post(url, params)
+
+        self.assert_200(response)
+        assert len(response.json()['results']) == 2  # only the entry of project that user is member
+
+        # try without authenticating the user
```
thenav56: `self.authenticate()` doesn't log out the user. So this comment is wrong.

Let's update this saying, try with default user which is a member in all projects.

apps/entry/filter_set.py
```diff
@@ -213,6 +219,12 @@ def authoring_organization_types_filter(self, qs, name, value):
             return qs.filter(organization_types__in=[ot.id for ot in value])
         return qs
 
+    def entry_ids_filter(self, qs, name, value):
+        value = eval(value)
```
thenav56: Let's not use `eval` due to security reasons.

`django_filters.ModelMultipleChoiceFilter` should be fine.

thenav56 (changes requested):

annshress (approved):

apps/entry/filter_set.py
```diff
@@ -213,6 +220,11 @@ def authoring_organization_types_filter(self, qs, name, value):
             return qs.filter(organization_types__in=[ot.id for ot in value])
         return qs
 
+    def entry_ids_filter(self, qs, name, value):
+        if value:
+            return qs.filter(id__in=[entry.id for entry in value])
+        return qs
+
```
thenav56: This might not be required. We should be able to define this in the `ModelMultipleChoiceFilter`.

https://django-filter.readthedocs.io/en/stable/ref/filters.html#to-field-name
https://django-filter.readthedocs.io/en/stable/ref/filters.html#lookup-expr

thenav56 (approved):

PR #743, opened 2021-04-21
Restrict same usergroup create multiple times in a project

k9845:
Fixes:
- #740 New Project Flow: Handle error while adding same user-group twice in a project

### DIFF
--- a/apps/project/serializers.py
+++ b/apps/project/serializers.py
@@ -459,19 +459,13 @@ class ProjectUserGroupSerializer(serializers.ModelSerializer):
         fields = '__all__'
         read_only_fields = ('project',)
 
-    def validate_project_usergroup(self, project, usergroup):
-        if ProjectUserGroupMembership.objects.filter(project=project, usergroup=usergroup).exists():
-            raise serializers.ValidationError({'usergroup': 'Usergroup already exist in the project'})
-
     def validate(self, data):
         data['project_id'] = int(self.context['view'].kwargs['project_id'])
         usergroup = data.get('usergroup')
-        usergroup = self.validate_project_usergroup(data['project_id'], usergroup)
+        if usergroup and ProjectUserGroupMembership.objects.filter(project=data['project_id'],
+                                                                   usergroup=usergroup).exists():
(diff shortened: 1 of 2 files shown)


apps/project/serializers.py
```diff
@@ -459,10 +459,20 @@ class Meta:
...
     def validate(self, data):
         data['project_id'] = int(self.context['view'].kwargs['project_id'])
+        usergroup = data.get('usergroup')
+        usergroup = self.validate_project_usergroup(data['project_id'], usergroup)
         return data
 
+    def create(self, validated_data):
+        instance, _ = ProjectUserGroupMembership.objects.get_or_create(**validated_data)
+        return instance
+
```
thenav56: This should be enough right?. Also for the patch method user group can be None.

```suggestion
    def validate(self, data):
        data['project_id'] = int(self.context['view'].kwargs['project_id'])
        usergroup = data.get('usergroup')
         if usergroup and ProjectUserGroupMembership.objects.filter(project=data['project_id'], usergroup=usergroup).exists():
            raise serializers.ValidationError({'usergroup': 'Usergroup already exist in the project'})
        return data
```

apps/project/tests/test_apis.py
```diff
@@ -538,6 +538,31 @@ def test_update_project_remove_ug(self):
...
+
+        # now try to create same usergroup from api level
+        data = {
+            'usergroup': self.ug1.id,
+            'role': self.normal_role.id
+        }
+        url = f'/api/v1/projects/{project.id}/project-usergroups/'
+        self.authenticate()
+        response = self.client.post(url, data)
+        self.assert_400(response)
```
thenav56, line 565: Let's add a retry after removing the membership we added manually.

thenav56 (changes requested):

apps/project/tests/test_apis.py
k9845: Done!

apps/project/serializers.py
k9845: Done!!

thenav56 (approved):

annshress (approved):

PR #744, opened 2021-04-23
Fix issue for membership role change

k9845:


apps/project/serializers.py
```diff
@@ -148,17 +148,21 @@ def validate_project(self, project):
...
         return project
 
-    def project_member_validation(self, project, member):
-        if ProjectMembership.objects.filter(project=project).filter(member=member).exists():
+    def project_member_validation(self, project, member, role):
+        if ProjectMembership.objects.filter(
+            project=project,
+            member=member,
+            role=role
+        ).exists():
```
thenav56: If a user tries to add (POST) another project-member using a different role, then there will be a 500 error. Let's add this to the test as well.

We don't need to update this. For POST this is fine. For update let's skip validation in validate as we don't need this.
Let's add test for:
- Add new project-membership. This should be okay.
- Update project-membership role. This should be okay.
- Add project-membership same as the one made before. This should raise an error.
    - Try again after removing the one made before. This should be okay.

annshress (approved):

thenav56 (approved):

PR #749, opened 2021-04-30
Migration for the project

k9845:


AdityaKhatri:
Closed because we went with another way of duplicating instance.

PR #750, opened 2021-04-30
Breaking: Rename analytical_statement to analytical_statements

tnagorra:
- Add uuid in analytical_entries
- Add uuid in analytical_statements
- Rename `analytical_statement` to `analytical_statements`
- Replace `uuid` with `client_id`
cc @AdityaKhatri  @subinasr


thenav56:
@tnagorra Both AnalyticalStatementEntry and AnalyticalStatement already have `cilent_id` field as they are inheriting the `UserResource` model. That 

https://github.com/the-deep/server/blob/9a84ece7e7883bf063282c366b810a9ec9823850/apps/user_resource/models.py#L5-L22

apps/analysis/models.py
```diff
@@ -63,6 +63,7 @@ class AnalyticalStatement(UserResource):
     )
     include_in_report = models.BooleanField(default=False)
     order = models.IntegerField()
+    uuid = models.CharField(max_length=255)
```
thenav56: Not required. Let's use the already defined field `client_id`.

apps/analysis/models.py
```diff
@@ -81,6 +82,7 @@ class AnalyticalStatementEntry(UserResource):
         on_delete=models.CASCADE
     )
     order = models.IntegerField()
+    uuid = models.CharField(max_length=255)
```
thenav56: Not required. Let's use the already defined field `client_id`

apps/analysis/serializers.py
```diff
@@ -21,7 +21,7 @@
 class AnlyticalEntriesSerializer(UserResourceSerializer):
     class Meta:
         model = AnalyticalStatementEntry
-        fields = ('id', 'order', 'entry')
+        fields = ('id', 'uuid', 'order', 'entry')
```
thenav56: This should work.
```suggestion
        fields = ('id', 'client_id', 'order', 'entry')
```

thenav56 (changes requested):
@tnagorra @k9845 Both AnalyticalStatementEntry and AnalyticalStatement already have `cilent_id` field as they are inheriting the `UserResource` model.

https://github.com/the-deep/server/blob/9a84ece7e7883bf063282c366b810a9ec9823850/apps/user_resource/models.py#L5-L22

apps/analysis/serializers.py
k9845: Done

thenav56 (approved):

annshress (approved):

Issue #757, opened 2021-05-13
Analysis Module: Create new entries API for pillar analysis that includes `discarded` boolean


AdityaKhatri:
* API should be similar to current entries filter API
* Add `discarded` boolean to filter entries that are discarded
* Add filter for discarded type: sending true will only send discarded entries, sending false will only send non-discarded entries
* Scope entries API under pillar analysis id

Issue #756, opened 2021-05-13
Analysis Module: Add discard entry action APIs


AdityaKhatri:
Entries of an analysis module can be tagged as discarded in:
1. Redundant
2. Too old
3. Anecdotal
4. Outlier

The client will send:
1. PillarAnaysisId
2. EntryId
3. DicardedType

Add un-discard entry API

PR #764, opened 2021-05-19
Add api for discarded and all entries in analysis pillar

k9845:
- Api for discarded entries : `/api/v1/analysis-pillar/{pillar.id}/discarded-entries/`
   perform post request as by passing `entry` and `tag`
   perform filter for tag as `/api/v1/analysis-pillar/{pillar.id}/discarded-entries/?tag={}`
- Api for the all entries: `/api/v1/analysis-pillar/{pillar.id}/entries/`
- Api for tags options `/api/v1/discardedentry-options/`

Related Issues:
- #756 
- #757


apps/analysis/admin.py
```diff
@@ -25,3 +26,8 @@ class AnalyticalStatementEntryInline(admin.TabularInline):
 @admin.register(AnalyticalStatement)
 class AnalyticalStatementAdmin(admin.ModelAdmin):
     inlines = [AnalyticalStatementEntryInline]
+
+
+@admin.register(DiscardedEntries)
+class DisCardedEntriesAdmin(admin.ModelAdmin):
```
thenav56:
```suggestion
class DiscardedEntriesAdmin(admin.ModelAdmin):
```

apps/analysis/filter_set.py
```diff
@@ -26,5 +29,17 @@ class AnalysisFilterSet(django_filters.FilterSet):
     )
 
     class Meta:
-        models = Analysis
+        model = Analysis
         fields = ()
+
+
+class DisCardedEntriesFilterSet(django_filters.FilterSet):
```
thenav56:
```suggestion
class DiscardedEntriesFilterSet(django_filters.FilterSet):
```

apps/analysis/models.py
```diff
@@ -49,6 +49,40 @@ def __str__(self):
...
+    TOO_OLD = 'too_old'
+    ANECDOTAL = 'anecdotal'
+    OUTLIER = 'outlier'
+
+    TAG_TYPES = (
+        (REDUNDANT, 'Redundant'),
+        (TOO_OLD, 'Too Old'),
+        (ANECDOTAL, 'ANECDOTAL'),
+        (OUTLIER, 'Outlier')
+    )
```
thenav56: Let's use django_enumfield from now on.

apps/analysis/models.py
```diff
@@ -49,6 +49,40 @@ def __str__(self):
...
+    entry = models.ForeignKey(
+        Entry,
+        on_delete=models.CASCADE
+    )
+    tag = models.CharField(
+        max_length=30,
+        choices=TAG_TYPES
+    )
+
```
thenav56, line 83: We will also need unique_together here for entry and analysis_pillar.

apps/analysis/views.py
```diff
@@ -113,6 +115,30 @@ def get_queryset(self):
         )
 
 
+class DiscardedEntriesViewSet(viewsets.ModelViewSet):
```
thenav56: Let's use AnalysisPillarDiscardedEntriesViewset
```suggestion
class AnalysisPillarDiscardedEntriesViewset(viewsets.ModelViewSet):
```

apps/analysis/views.py
```diff
@@ -113,6 +115,30 @@ def get_queryset(self):
...
+    serializer_class = DiscardedEntriesSerializer
+    permissions_classes = [permissions.IsAuthenticated]
+    filterset_class = DisCardedEntriesFilterSet
```
thenav56: Make sure permissions are check here.

apps/analysis/views.py
```diff
@@ -113,6 +115,30 @@ def get_queryset(self):
...
+        }
+
+
+class PillarEntriesViewSet(EntryViewSet):
+
+    def get_queryset(self):
+        discarded_entries = DiscardedEntries.objects.filter(
+            analysis_pillar=self.kwargs['analysis_pillar_id']
+        ).values('entry')
+        return Entry.objects.exclude(id__in=discarded_entries)
```
thenav56: Make sure permissions are handled here. This might take care of the permission. But let's add a test to validate that too.
```suggestion
        return super().get_queryset().exclude(id__in=discarded_entries)
```

apps/analysis/views.py
```diff
@@ -113,6 +115,30 @@ def get_queryset(self):
...
+    def get_serializer_context(self):
+        return {
+            **super().get_serializer_context(),
+            'analysis_pillar_id': self.kwargs.get('analysis_pillar_id'),
+        }
+
+
+class PillarEntriesViewSet(EntryViewSet):
+
+    def get_queryset(self):
```
thenav56, line 136: We will need a dynamic filter here instead of just discarded_entries.

thenav56 (changes requested):
Let's also update the PR description by including related tickets.

apps/analysis/views.py
```diff
@@ -113,6 +114,33 @@ def get_queryset(self):
...
+        queryset = super().get_queryset()
+        filters = self.get_entry_fiters()
+        discard_entires = filters.get('discarded')
+        analysis_pillar_id = self.kwargs['analysis_pillar_id']
+        discarded_entries = DiscardedEntry.objects.filter(analysis_pillar=analysis_pillar_id).values('entry')
+        if discard_entires:
+            return queryset.filter(id__in=discarded_entries)
+        return queryset.exclude(id__in=discarded_entries)
+
```
annshress: There is good amount of typos in here.

- `discard_entires` vs `discarded_entries`
- `get_entry_fiters()`

thenav56: Done. https://github.com/the-deep/server/pull/764/commits/acb48a770fe0bf332c5775d0aaf365ebe69cbc8a

annshress (approved):

apps/analysis/tests/test_apis.py
```diff
@@ -583,3 +584,99 @@ def test_analysis_overview_in_project(self):
...
+        data = {
+            'entry': entry.id,
+            'tag': DiscardedEntry.TagType.REDUNDANT
+        }
+        url = f'/api/v1/analysis-pillar/{pillar1.id}/discarded-entries/'
+        self.authenticate(user)
+        response = self.client.post(url, data)
+        self.assert_201(response)
+        self.assertEqual(response.data['analysis_pillar'], pillar1.id)
+        self.assertEqual(response.data['entry'], entry.id)
```
thenav56, line 605: Also, users shouldn't be able to discard an entry to which they don't have permission too.

thenav56 (changes requested):

annshress (approved):

apps/analysis/serializers.py
```diff
@@ -52,6 +54,25 @@ def validate(self, data):
...
+
+    class Meta:
+        model = DiscardedEntry
+        fields = '__all__'
+
+    def validate(self, data):
+        if data['analysis_pillar'].id != int(self.context['view'].kwargs.get('analysis_pillar_id', None)):
+            raise serializers.ValidationError(
+                f'Analysis pillar doesnot match'
+            )
```
thenav56: We were directly using analysis_pillar_id, right? Why are we asking this from the client and validating it?

thenav56 (changes requested):

apps/analysis/serializers.py
k9845: During validation we cannot get the `analysis_pillar` from the url kwargs since we were passing the `analysis_pillar_id`, we require `analysis_pillar` project for the validation.

thenav56: We can get the analysis_pillar using analysis_pillar_id. There is no need to ask for id in both URL and body from the client.

k9845: Done `8d74dae`

thenav56 (approved):

PR #748, opened 2021-04-29
Upgrade packages

bimal125:
Following are main changes in this repo:

- Upgraged django along with other packages
- Setup graphene
- Polish tests

PR #788, opened 2021-06-03
Bare mimimum analysis framework graphql endpoint

annshress:
Merge after https://github.com/the-deep/server/pull/748. Currently the destination branch is the same, to ease in review.

This PR will:

- [x] Setup base for Analysis Framework CRUD
- [x] Port mutation related rest api tests
- [x] Add generic mutation for create/update/delete with existing permission classes in mind

What it does not contain:

- [ ] Dataloaders for queries (We found the current dataloader in helix cannot handle [aliases](https://graphqlmastery.com/blog/how-to-use-graphql-aliases))
- [ ] Full fledged READ like there was in rest api (Because we are anticipating good amount of change in schema in both client and server here)

@samshara @suraj-ram-shrestha


apps/analysis_framework/tests/test_mutations.py
```diff
@@ -0,0 +1,201 @@
...
+        response = self.query(
+            self.create_mutation,
+            input_data=self.input
+        )
+        content = response.json()
+        self.assertIsNotNone(content['errors'][0]['message'])
+        # The actual message is
+        # You don't have permissions to create private framework
+        self.assertIn('permission', content['errors'][0]['message'])
+
```
k9845: ain't these tests the same?

k9845 (approved):
LGTM

thenav56 (approved):

Issue #798, opened 2021-06-21
Add new api for geoarea


k9845:
## Acceptance Criteria
- Should support pagination
- Response should be flat
- Support for search

PR #799, opened 2021-06-21
Add api for geo-area

k9845:
Addresses #798

## Changes
- Support pagination
- Support for search
- Make response flat
- Api `/api/v1/projects/{project.id}/geo-area/`

@AdityaKhatri 

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/geo/filter_set.py
+++ b/apps/geo/filter_set.py
@@ -1,0 +1,18 @@
+import django_filters
+
+from geo.models import GeoArea
+
+class GeoAreaFilterSet(django_filters.rest_framework.FilterSet):
+    label = django_filters.CharFilter(
+        label='Geo Area Label',
+        method='geo_area_label'
+    )
+
+    class Meat:
+        model = GeoArea
+        fields = ()
+
+    def geo_area_label(self, queryset, name, value):
+        if value:
+            return queryset.filter(label__icontains=value)
+        return queryset
--- a/apps/geo/views.py
+++ b/apps/geo/views.py
@@ -15,7 +15,10 @@ from rest_framework import (
 from rest_framework.decorators import action
 import django_filters
 
-from deep.permissions import ModifyPermission
+from deep.permissions import (
+    ModifyPermission,
+    IsProjectMember
+)
 from user_resource.filters import UserResourceFilterSet
 
 from project.models import Project
@@ -25,7 +28,7 @@ from .serializers import (
     RegionSerializer,
     GeoAreaSerializer
 )
-
+from .filter_set import GeoAreaFilterSet
 from geo.tasks import load_geo_areas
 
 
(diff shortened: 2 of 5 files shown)


annshress (approved):

deep/urls.py
```diff
@@ -251,6 +252,8 @@
                 basename='region')
 router.register(r'admin-levels', AdminLevelViewSet,
                 basename='admin_level')
+router.register(r'geo-area', GeoAreaView,
+                basename='geo_area')
```
thenav56: Let's scope this under project.

apps/geo/views.py
```diff
@@ -241,5 +242,43 @@ def get(self, request, version=None):
...
+                admin_level__region__project=project
+            )
+        # this is added for the serch purpose
+        label = self.request.query_params.get('label')
+        if label:
+            return queryset.filter(
+                models.Q(title__icontains=label) |
+                models.Q(admin_level__title=label)
+            )
+        return queryset
```
thenav56: Let's remove this.

apps/geo/views.py
```diff
@@ -241,5 +242,43 @@ def get(self, request, version=None):
...
+            admin_level_title=models.F('admin_level__title'),
+            key=models.F('id')
+        )
+        project = self.request.query_params.get('project')
+        if project:
+            project = Project.objects.get(id=project)
+            if not project.is_member(self.request.user):
+                raise exceptions.PermissionDenied()
+            return queryset.filter(
+                admin_level__region__project=project
```
thenav56: For filter let's use search filter backend or django_filter. It will be easier if we need to add additional fields and also for documentation.

apps/geo/views.py
```diff
@@ -241,5 +242,43 @@ def get(self, request, version=None):
...
+
+
+class GeoAreaView(viewsets.ReadOnlyModelViewSet):
+    permissions_classes = [permissions.IsAuthenticated]
+    serializer_class = GeoAreaSerializer
+
+    def get_queryset(self):
+        queryset = GeoArea.objects.prefetch_related(
+            'admin_level',
+            'admin_level__region'
```
thenav56: We don't need prefetch_related if we are getting the values from annotation.

Also, Note:
- Prefetch related is used if there are multiple objects for that field (this will separately prefetch in bulk)
- We use select related for single objects. (this will do join)

apps/geo/views.py
```diff
@@ -241,5 +242,43 @@ def get(self, request, version=None):
...
+                models.F('admin_level__title'),
+                models.Value('/'),
+                models.F('title'),
+                output_field=models.fields.CharField()
+            ),
+            region=models.F('admin_level__region_id'),
+            region_title=models.F('admin_level__region__title'),
+            admin_level_level=models.F('admin_level__level'),
+            admin_level_title=models.F('admin_level__title'),
+            key=models.F('id')
```
thenav56, line 270: Are these fields enough for the client-side? @AdityaKhatri

thenav56 (changes requested):

apps/geo/filter_set.py
```diff
@@ -0,0 +1,20 @@
...
+
+    class Meat:
+        model = GeoArea
+        fields = ()
+
+    def geo_area_label(self, queryset, name, value):
+        if value:
+            return queryset.filter(
+                label__icontains=value
+            )
```
thenav56:
```suggestion
            return queryset.filter(label__icontains=value)
```

apps/geo/views.py
```diff
@@ -241,5 +245,27 @@ def get(self, request, version=None):
...
+
+class GeoAreaView(viewsets.ReadOnlyModelViewSet):
+    permission_classes = [permissions.IsAuthenticated, IsProjectMember]
+    serializer_class = GeoAreaSerializer
+    filterset_class = GeoAreaFilterSet
+
+    def get_queryset(self):
+        return GeoArea.objects.filter(
+            admin_level__region__project=self.kwargs['project_id']
+        ).distinct().annotate(
```
thenav56: let's use `distinct()` at the end. Also, we might not need distinct here (we need distinct when we join with many rows).

thenav56 (changes requested):

annshress (approved):

thenav56 (approved):

PR #803, opened 2021-06-22
Create api to clone analyis-pillar

k9845:
Addresses #802 

## Changes
- Cloning the analytical statement along with the analysis-pillars
- New API for the analysis-pillar clone
- Api `/api/v1/projects/{project.id}/analysis/{analysis.id}/pillars/{pillar.id}/clone-pillar/`
@AdityaKhatri 

- [x] temporary files, auto-generated files, or secret keys
- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [] translations

### DIFF
--- a/apps/analysis/models.py
+++ b/apps/analysis/models.py
@@ -51,12 +51,21 @@ class AnalysisPillar(UserResource):
 
     def clone_pillar(self, title):
         pillar_cloned = copy.deepcopy(self)
+
+        def _get_clone_ready(obj, pillar_cloned):
+            obj.client_id = None
+            obj.pk = None
+            obj.analysis_pillar = pillar_cloned
+            return obj
+
         pillar_cloned.pk = None
         pillar_cloned.client_id = None
-        pillar_cloned.title = f'{title} (cloned)'
+        pillar_cloned.title = title
         pillar_cloned.save()
-        [statement.clone_to(pillar_cloned) for statement
-        in self.analyticalstatement_set.all()]
+        analytical_statements = self.analyticalstatement_set.all()
+        AnalyticalStatement.objects.bulk_create([
+            _get_clone_ready(analytical_statement, pillar_cloned) for analytical_statement in analytical_statements
+        ])
         return pillar_cloned
 
 
@@ -115,16 +124,6 @@ class AnalyticalStatement(UserResource):
     def __str__(self):
         return self.statement and self.statement[:255]
 
-    def clone_to(self, analysis_pillar):
-        cloned_statement = copy.deepcopy(self)
-        cloned_statement.pk = None
-        cloned_statement.client_id = None
-        cloned_statement.statement = self.statement
-        cloned_statement.analysis_pillar = analysis_pillar
-        cloned_statement.save()
-        [statement_entry.clone_to(cloned_statement) for statement_entry
-        in self.analyticalstatemententry_set.all()]
-        return cloned_statement
 
 class AnalyticalStatementEntry(UserResource):
     entry = models.ForeignKey(
@@ -140,11 +139,3 @@ class AnalyticalStatementEntry(UserResource):
     class Meta:
         ordering = ('order',)
         unique_together = ('entry', 'analytical_statement')
-
-    def clone_to(self, analytical_statement):
-        cloned_statement_entry = copy.deepcopy(self)
-        cloned_statement_entry.pk = None
-        cloned_statement_entry.client_id = None
-        cloned_statement_entry.analytical_statement = analytical_statement
-        cloned_statement_entry.save()
-        return cloned_statement_entry
--- a/apps/analysis/models.py
+++ b/apps/analysis/models.py
@@ -49,6 +49,14 @@ class AnalysisPillar(UserResource):
     def __str__(self):
         return self.title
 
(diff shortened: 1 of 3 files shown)


apps/analysis/views.py
```diff
@@ -114,6 +114,29 @@ def get_queryset(self):
...
 
+    @action(
+        detail=True,
+        url_path='clone-pillar',
+        permission_classes=[IsProjectMember],
+        methods=['post']
+    )
+    def clone_pillar(self, request, project_id, analysis_id, pk=None, version=None):
+        pillar = self.get_object()
+        cloned_title = request.data.get('title').strip()
```
thenav56, line 125: `cloned_title` is not used?

apps/analysis/models.py
```diff
@@ -50,6 +50,14 @@ class AnalysisPillar(UserResource):
     def __str__(self):
         return self.title
 
+    def clone_pillar(self):
+        pillar_cloned = copy.deepcopy(self)
+        pillar_cloned.pk = None
+        pillar_cloned.client_id = None
+        pillar_cloned.title = f'{self.title} (cloned)'
+        pillar_cloned.save()
+        return pillar_cloned
```
thenav56: Is this behavior okay? We don't need to clone statements too?

thenav56 (changes requested):

apps/analysis/tests/test_apis.py
```diff
@@ -423,6 +423,29 @@ def test_clone_analysis(self):
...
+        self.assert_201(response)
+
+        self.assertNotEqual(response.data['id'], pillar.id)
+        self.assertEqual(response.data['title'], f'{pillar.title} (cloned)')
+
+        # authenticating with user that is not project member
+        self.authenticate(user2)
+        response = self.client.post(url, data)
+        self.assert_403(response)
+
```
annshress, line 465: Lets add some analysis statements to the pillar and check if they are cloning as well. And test accordingly if we need to actually clone the inner statements or not.

apps/analysis/models.py
```diff
@@ -50,6 +50,16 @@ class AnalysisPillar(UserResource):
...
         return self.title
 
+    def clone_pillar(self, title):
+        pillar_cloned = copy.deepcopy(self)
+        pillar_cloned.pk = None
+        pillar_cloned.client_id = None
+        pillar_cloned.title = f'{title} (cloned)'
+        pillar_cloned.save()
+        [statement.clone_to(pillar_cloned) for statement
+        in self.analyticalstatement_set.all()]
```
thenav56: Maybe use bulk_create here?

apps/analysis/models.py
```diff
@@ -106,6 +116,16 @@ class Meta:
...
 
+    def clone_to(self, analysis_pillar):
+        cloned_statement = copy.deepcopy(self)
+        cloned_statement.pk = None
+        cloned_statement.client_id = None
+        cloned_statement.statement = self.statement
+        cloned_statement.analysis_pillar = analysis_pillar
+        cloned_statement.save()
+        [statement_entry.clone_to(cloned_statement) for statement_entry
+        in self.analyticalstatemententry_set.all()]
```
thenav56: @AdityaKhatri 
Do we also need to clone entries as well? Option to skip this as well?

@k9845 
Let's not use list comprehension here, `for` loop is fine. List comprehension is fine if we are using bulk_create.

apps/analysis/views.py
```diff
@@ -114,6 +114,31 @@ def get_queryset(self):
...
+        if not cloned_title:
+            raise exceptions.ValidationError({
+                'title': 'Title should be present',
+            })
+        new_pillar = pillar.clone_pillar(
+            cloned_title
+        )
```
thenav56:
```suggestion
        new_pillar = pillar.clone_pillar(cloned_title)
```

apps/analysis/models.py
```diff
@@ -50,6 +50,16 @@ class AnalysisPillar(UserResource):
     def __str__(self):
         return self.title
 
+    def clone_pillar(self, title):
+        pillar_cloned = copy.deepcopy(self)
+        pillar_cloned.pk = None
+        pillar_cloned.client_id = None
+        pillar_cloned.title = f'{title} (cloned)'
```
thenav56: Do we need to append `(cloned)` to the user-provided title? @AdityaKhatri

thenav56 (changes requested):

apps/analysis/models.py
AdityaKhatri: Can we have the logic where if its the same name we append the cloned status?

thenav56: Let's not append `(cloned)`. Since we need to create/store the cloned_from value, the client can show if it was cloned or not.

k9845: Updated in `fa0ef22`

apps/analysis/tests/test_apis.py
k9845: added `fa0ef22`

k9845:
Used different approach with cloning the analysis instead of analysis pillar

Issue #809, opened 2021-06-27
New project flow: Add `published` status to regions


AdityaKhatri:
![image.png](https://zube.io/files/deep/869019d18f538b57c2869db37703d1cc-image.png)

## Changes

* Geoareas(Entry option) will only appear during tagging if a region is published
* Migration should add `published: true` for all previous regions
* Once published cannot be modified or unplublished.

PR #832, opened 2021-07-01
Add published boolean field to region model

k9845:
Addresses
- #809 

## Changes
- Added published boolean field in Region model
- Added api for the publish `/api/v1/regions/{region.id}/publish/`

@samshara 

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations


apps/geo/migrations/0038_region_published.py
```diff
@@ -0,0 +1,18 @@
...
+        ('geo', '0037_auto_20210503_0431'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='region',
+            name='published',
+            field=models.BooleanField(default=True),
+        ),
+    ]
```
thenav56: Let's merge this two migration to one.

apps/geo/serializers.py
```diff
@@ -64,6 +64,11 @@ def validate_project(self, project):
             raise serializers.ValidationError('Invalid project')
         return project.id
 
+    def validate(self, data):
+        if self.instance and self.instance.published == True:
```
thenav56:
```suggestion
        if self.instance and self.instance.is_published:
```

Also, add `is_published` to Meta.exclude. And add a new endpoint to publish the region.

apps/geo/serializers.py
```diff
@@ -64,6 +64,11 @@ def validate_project(self, project):
...
+            raise serializers.ValidationError({'published': 'Published once set True cannot be revert back'})
```
thenav56:
```suggestion
            raise serializers.ValidationError('Published region can\'t be changed. Please contact Admin')
```
cc: @AdityaKhatri

apps/geo/models.py
```diff
@@ -111,7 +114,7 @@ def can_modify(self, user):
                 project__regions=self,
                 member=user,
                 role__in=ProjectRole.get_admin_roles(),
-            ).exists())
+            ).exists())) and self.published == False
```
thenav56:
```suggestion
            ).exists())) and not self.is_published
```

apps/geo/models.py
```diff
@@ -19,6 +21,7 @@ class Region(UserResource):
     code = models.CharField(max_length=10)
     title = models.CharField(max_length=255)
     public = models.BooleanField(default=True)
+    published = models.BooleanField(default=False)
```
thenav56: Additional work to do:
- Let's use `is_published`.
- `is_published` should be true for all current regions (we need to make sure that current regions are published)
- For new regions, `is_published` should be False by default and can be set by hitting an endpoint.

thenav56 (changes requested):

thenav56:
@k9845 Add an endpoint to publish a region.
Also, tests are failing.

cc: @AdityaKhatri @suraj-ram-shrestha

apps/geo/serializers.py
AdityaKhatri: Looks good to me

apps/geo/models.py
```diff
@@ -111,9 +115,14 @@ def can_modify(self, user):
                 project__regions=self,
                 member=user,
                 role__in=ProjectRole.get_admin_roles(),
-            ).exists())
+            ).exists())) and not self.is_published
```
thenav56: Also, move not self.is_published check to the top.

apps/geo/models.py
```diff
@@ -111,9 +115,14 @@ def can_modify(self, user):
...
         )
 
+    def can_publish(self, user):
+        if not Region.objects.filter(id=self, created_by=user):
+            raise ValidationError("Can be published by the user who created it")
+        return user
```
thenav56: Why is this here?

apps/geo/serializers.py
```diff
@@ -50,7 +50,7 @@ class RegionSerializer(RemoveNullFieldsMixin,
 
     class Meta:
         model = Region
-        exclude = ('geo_options',)
+        exclude = ('geo_options', 'is_published')
```
thenav56: On second thought, remove is_published from here and add it as a read-only boolean field in the serializer.

apps/geo/views.py
```diff
@@ -96,6 +99,23 @@ def get_intersects(self, request, pk=None, version=None):
             for id, geom in geoms
         ])
 
+    @action(
+        detail=True,
+        url_path='published',
```
thenav56: `publish` makes more sense.

```suggestion
        url_path='publish',
```

apps/geo/views.py
```diff
@@ -96,6 +99,23 @@ def get_intersects(self, request, pk=None, version=None):
...
+        methods=['post'],
+        serializer_class = RegionSerializer,
```
thenav56: Flake8, please check.

```suggestion
        serializer_class=RegionSerializer,
```

apps/geo/views.py
```diff
@@ -96,6 +99,23 @@ def get_intersects(self, request, pk=None, version=None):
...
+        permission_classes=[permissions.IsAuthenticated]
+    )
+    def get_published(self, request, pk=None, version=None):
+        region = self.get_object()
+        if not Region.objects.filter(id=region.id, created_by=request.user):
+            raise ValidationError('Can be published by user who created it')
+
+        region.is_published = True
+        region.save()
```
thenav56: Good habit to use updated_fields when we are changing single field only.

```suggestion
        region.save(update_fields=['is_published'])
```

thenav56 (changes requested):

apps/geo/serializers.py
```diff
@@ -47,7 +47,7 @@ class RegionSerializer(RemoveNullFieldsMixin,
         write_only=True,
         required=False,
     )
-
+    is_published = serializers.BooleanField(read_only=True)
```
thenav56, line 50: Flake8.  :expressionless: 

```suggestion
    is_published = serializers.BooleanField(read_only=True)

```

apps/geo/models.py
```diff
@@ -111,9 +115,12 @@ def can_modify(self, user):
                 project__regions=self,
                 member=user,
                 role__in=ProjectRole.get_admin_roles(),
-            ).exists())
+            ).exists()))
         )
 
+    def can_publish(self, user):
+        return self.created_by == user
```
thenav56: Easy to read.

```suggestion
        return user == self.created_by
```

thenav56 (changes requested):

thenav56 (dismissed):

AdityaKhatri:
Lets merge this a bit later after we diverge as this is not something we'll be implementing in the old design.

annshress (dismissed):

thenav56 (approved):

Issue #817, opened 2021-06-28
Create graphql query and mutation for usergroups


tnagorra:
### Mutation

- update usergroup information (to update one's own profile)
* Add a user to usergroup
* Remove a user to usergroup

### Query

- get /usergroups/
- get /usergroup/

>NOTE: "Modify user permission" will be handled in next sprint

PR #836, opened 2021-07-05
Feature/migrate to pipenv

thenav56:
Addresses
- #818
- #819
- #820
- #821
- #817 
- #844 

## Changes

- Change dockerfile (WARNING)
    - Use python buster image as base image.
    - Remove chrome/ploty (Tabular, thumbnail feature will not work) (~2.5GB size reduction)
        - Add serverless module for this feature in future.
    - Add pipenv
        - Upgrade all the packages.
- Move requirements.txt to requirements.old.txt

- Add factory-boy.
- Refactor captcha logic. (Add Tests)

- Add mutation for (with test)
    - Login/Logout #818
    - Register #819
    - changePassword #820
    - resetPassword #821
    - createUserGroup, updateUserGroup, deleteUserGroup #817

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

## Note (WARNING)
Merging this to develop will have these major changes:
- We need to build a server after each Pipfile changes.
- We need to set up pipenv/pyenv in local to add/remove packages. [Follow this](https://outline.togglecorp.com/share/37273873-8452-4c93-871a-76c169c6f6b0) cc: @annshress @k9845 @bimal125 
- Tabular Viz and Thumbnail generation will not work. (Due to docker image size reduction)


apps/user/schema.py
```diff
@@ -0,0 +1,60 @@
...
+        model = User
+        fields = (
+            'id', 'email', 'first_name', 'last_name', 'is_active'
+        )
+
+    display_name = graphene.String()
+
+    @staticmethod
+    def resolve_display_name(root, info, **kwargs):
+        return root.profile.get_display_name()
```
k9845: Can we use a property decorator and define it somewhere in models?

apps/user/schema.py
```diff
@@ -0,0 +1,60 @@
...
+        model = User
+        fields = (
+            'id', 'email', 'first_name', 'last_name',
+            'is_active', 'last_login',
+        )
+
+
+# XXX: Is this okay to do?
+class UserListType(CustomDjangoListObjectType):
+    base_type = UserType
```
k9845: Is this possible?

thenav56: Yup. @annshress Is this okay to do?

https://github.com/the-deep/server/pull/836/files#diff-ac5f6b70238d8921b24e3faace6afd8a43c50a480f3f5257d94d45c4c2e0444aR126

apps/user/schema.py
thenav56: This is defined in the model but in the profile. I am thinking of using the abstract model in deep too (to avoid additional profile join).

annshress (approved):

apps/user/schema.py
thenav56: This is removed.

apps/user/schema.py
thenav56: Using property.

k9845 (approved):

thenav56:
Opened a new PR with branch as [project/graphql](https://github.com/the-deep/server/tree/project/graphql)
https://github.com/the-deep/server/pull/887

PR #864, opened 2021-07-14
Fix django filter issues after django version upgrade

k9845:
Addresses 
 - When filtered by Created date filter, getting 400 Bad Request error(#https://zube.io/deep/deep/c/3509)
 - When filtered by Assignee, we are facing 500 Internal Server Error(#https://zube.io/deep/deep/c/3508)
 - In Assessments form, when searched by Created By proper list is not been displayed(#https://zube.io/deep/deep/c/3510)

## Changes
- Fixed Django filter issue after version upgrade

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/ary/tests/test_apis.py
+++ b/apps/ary/tests/test_apis.py
@@ -288,7 +288,7 @@ class AssessmentTests(TestCase):
         self.update_obj(self.create(Assessment, lead=lead3, project=project), created_at=now + relativedelta(days=-2))
         self.update_obj(self.create(Assessment, lead=lead4, project=project), created_at=now)
 
-        params = {'created_at__gt': now.strftime('%Y-%m-%d%z')}
+        params = {'created_at__gte': now.strftime('%Y-%m-%d%z')}
         url = f'/api/v1/assessments/'
         self.authenticate()
         respose = self.client.get(url, params)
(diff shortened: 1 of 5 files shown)


apps/lead/tests/test_apis.py
```diff
@@ -1028,6 +1028,27 @@ def test_lead_filter(self):
...
+        self.authenticate()
+        response = self.client.get(url)
+        assert len(response.data['results']) == 2
+
+        # filter by user who is not assignee in any of the lead
+        url = f'/api/v1/leads/?assignee={user3.id}'
+
+        self.authenticate()
+        response = self.client.get(url)
+        assert len(response.data['results']) == 0
```
thenav56: This will be more readable.

```suggestion
        # authenticate user
        self.authenticate()
        
        # filter by user who is assignee in some leads
        url = f'/api/v1/leads/?assignee={user1.id}'
        response = self.client.get(url)
        assert len(response.data['results']) == 2

        # filter by user who is not assignee in any of the lead
        url = f'/api/v1/leads/?assignee={user3.id}'
        response = self.client.get(url)
        assert len(response.data['results']) == 0
```

apps/ary/tests/test_apis.py
```diff
@@ -269,3 +273,55 @@ def test_ary_copy(self):
...
+        response = self.client.get(url)
+        self.assert_200(response)
+        self.assertIn(project2.id, [item['key'] for item in response.data['project']])
+        self.assertIn(project2.title, [item['value'] for item in response.data['project']])
+        # gives the member of the project
+        self.assertIn(user1.id, [item['key'] for item in response.data['created_by']])
+        self.assertEqual(
+            set([item['key'] for item in response.data['created_by']]),
+            set([user1.id, user2.id])
+        )
```
thenav56: - Let's user assertEqual + set for project as well
- we don't need `self.assertIn(user1.id, [item['key'] for item in response.data['created_by']])`

Let's also add check from the user2 perspective as well. `self.authenticate(user2)`

thenav56 (changes requested):

apps/ary/tests/test_apis.py
thenav56: Looks like the current implementation is incorrect. Let's do this
- `created_by` should be generated using assessment created_by (filter by projects)
- projects are recalculated for `option['project']`, let's remove that as well. (use the projects as calculated at the start of the view)

apps/ary/views.py
```diff
@@ -166,28 +168,19 @@ def get(self, request, version=None):
...
+            key=models.F('created_by'),
+            value=models.F('created_by__username')
+        ).values('key', 'value')
+
+        options['project'] = [
+            {
+                'key': project.id,
+                'value': project.title,
+            } for project in projects.distinct()
+        ] if (fields is None or 'project' in fields) else []
```
thenav56: Let's do this way (using get_display_name). Will do query optimization in graphql.
```suggestion
        if (fields is None or 'created_by' in fields):
            assessment_qs = Assessment.objects.filter(project__in=projects)
            options['created_by'] = [
                {
                    'key': user.id,
                    'value': user.profile.get_display_name(),
                }
                for user in User.objects.filter(pk__in=assessment_qs.distinct().values('created_by')).select_related('profile')
            ]

        if (fields is None or 'project' in fields):
            options['project'] = [
                {
                    'key': project.id,
                    'value': project.title,
                } for project in projects.distinct()
            ]
```

apps/ary/tests/test_apis.py
```diff
@@ -269,3 +273,73 @@ def test_ary_copy(self):
...
+        self.create(Assessment, lead=lead3, project=project1, created_by=user2)
+
+        # filter by project2
+        url = f'/api/v1/assessment-options/?project={project2.id}'
+        self.authenticate(user1)
+        response = self.client.get(url)
+        self.assert_200(response)
+        self.assertIn(project2.id, [item['key'] for item in response.data['project']])
+        self.assertIn(project2.title, [item['value'] for item in response.data['project']])
+        self.assertEqual(len(response.data['project']), 1)
```
thenav56:
```suggestion
        projects = response.data['project']
        self.assertEqual(len(projects), 1)
        self.assertEqual(projects[0]['key'], project2.id, projects)
        self.assertEqual(projects[0]['value'], project2.title, projects)
```

thenav56 (changes requested):

thenav56 (approved):

AdityaKhatri:
created_at filter is not working

Issue #841, opened 2021-07-08
Leads: Add `is_assessment` field in leads


AdityaKhatri:
### New Addition
* Add `is_assessment_lead` boolean field to lead model
* If a lead already has assessment attached to it, user cannot  make `is_assessment_lead` false (send user readable error message)

### Changes to old modules/configurations
* Old leads that have assessment in them should be migrated to have `is_assessment_lead` as true
* Old leads that do not have assessment in them should be migrated to have `is_assessment_lead` as false

PR #865, opened 2021-07-15
Add boolean `is_assessment_lead` in lead model

k9845:
Addresses https://github.com/the-deep/server/issues/841

## Changes
 - Added `is _assessment_lead` boolean to lead model
 - Migrated all the leads that have migrations to `is _assessment_lead=True` 

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations


apps/lead/migrations/0037_auto_20210715_0432.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def set_lead_is_assessment_(Lead, Assessment):
+    assessments = Assessment.objects.all()
+    leads = Lead.objects.filter(
+        assessment__in=assessments
+    )
+    for lead in leads:
+        lead.is_assessment_lead = True
+        lead.save(update_fields=['is_assessment_lead'])
```
thenav56: We should be able to use just .update() here.

Also, let's use `_set_lead_is_assessment`

apps/lead/serializers.py
```diff
@@ -216,6 +216,10 @@ def validate(self, data):
             data, self.instance,
             File.objects.filter(pk=attachment_id).first()
         )
+        if self.instance and self.instance.is_assessment_lead:
+            raise serializers.ValidationError(
+                {'is_assessment_lead': 'Lead Assessment once set can\'t be changed. Please contact Admin'}
+            )
```
thenav56: We need to check if the lead already has an assessment before making it false. So,
- Always allow setting True.
- For setting False make sure there is no assessment attached to that lead.

thenav56 (changes requested):

apps/lead/migrations/0037_auto_20210715_0432.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2 on 2021-07-15 04:32
+
+from django.db import migrations
+
+
+def _set_lead_is_assessment(Lead, Assessment):
+    assessments = Assessment.objects.all()
+    Lead.objects.filter(
+        assessment__in=assessments
+    ).update(is_assessment_lead=True)
```
thenav56:
```suggestion
def _set_lead_is_assessment(Lead):
    # Set is_assessment_lead True for lead already having assessment.
    Lead.objects.filter(assessment__isnull=False).update(is_assessment_lead=True)
```

apps/lead/serializers.py
```diff
@@ -210,6 +211,11 @@ def add_update__validate(data, instance, attachment=None):
             attachment,
         )
 
+    def validate_is_assessment_lead(self, value):
+        if hasattr(self.instance, 'assessment'):
+            raise serializers.ValidationError('Lead Assessment once set can\'t be changed. Please contact Admin')
+        return value
```
thenav56:
```suggestion
    def validate_is_assessment_lead(self, value):
        # Allow setting True
        # For False make sure there are no assessment attached.
        if value is False and hasattr(self.instance, 'assessment'):
            raise serializers.ValidationError('Lead already has an assessment.')
        return value
```

thenav56 (changes requested):

thenav56 (dismissed):

bimal125 (dismissed):

thenav56 (dismissed):

PR #867, opened 2021-07-19
Check entries otptions for entry filter

k9845:
Addresses
- Entry filter not working(https://zube.io/deep/deep/c/3647)
- Added by not appearing in project-usergroups(https://zube.io/deep/deep/c/3723)

## Changes

* Detailed list or prose of changes
* Breaking changes
* Changes to configurations

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/project/serializers.py
+++ b/apps/project/serializers.py
@@ -471,6 +471,12 @@ class ProjectUserGroupSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError({'usergroup': 'Usergroup already exist in the project'})
         return data
 
+    def create(self, validated_data):
+        user_group = super().create(validated_data)
+        user_group.added_by = self.context['request'].user
+        user_group.save()
+        return user_group
+
 
 class ProjectRecentActivitySerializer(serializers.Serializer):
     id = serializers.IntegerField()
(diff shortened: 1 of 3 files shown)


apps/project/serializers.py
```diff
@@ -471,6 +471,12 @@ def validate(self, data):
             raise serializers.ValidationError({'usergroup': 'Usergroup already exist in the project'})
         return data
 
+    def create(self, validated_data):
+        user_group = super().create(validated_data)
+        user_group.added_by = self.context['request'].user
+        user_group.save()
```
thenav56: Let's always use **update_fields** when we are updating one field.

apps/project/serializers.py
```diff
@@ -471,6 +471,12 @@ def validate(self, data):
             raise serializers.ValidationError({'usergroup': 'Usergroup already exist in the project'})
         return data
 
+    def create(self, validated_data):
+        user_group = super().create(validated_data)
```
thenav56: It's not user_group, it should be project_user_group_membership or just membership.

thenav56 (changes requested):

thenav56 (approved):

Issue #848, opened 2021-07-13
Graphql: Add lead query


suraj-ram-shrestha:
## Acceptance Criteria
*  Add lead query

>NOTE: Refer to mockup for the table view.

Issue #847, opened 2021-07-13
Graphql: Add a query for project


suraj-ram-shrestha:
## Acceptance Criteria
* Add a query for the project

Issue #845, opened 2021-07-12
Leads: Bulk lead save mutation


suraj-ram-shrestha:
### New Addition
- Support individual atomic.
- Valid validation error handling.

PR #868, opened 2021-07-19
Add Project + lead Query

thenav56:
Addresses
- #848
- #847
- #845 
## Changes

- Add Project + lead Query
- Setup basic permissions for Graphql.
- Using poetry. (Remove pipenv)
- Add lead mutations.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests


apps/lead/schema.py
```diff
@@ -0,0 +1,93 @@
...
+    class Meta:
+        model = Lead
+        fields = (
+            'id', 'title', 'created_by', 'created_at', 'modified_by', 'modified_at',
+            'project', 'lead_group', 'assignee', 'published_on',
+            'source_type', 'priority', 'confidentiality', 'status',
+            'text', 'url', 'website', 'attachment', 'emm_entities'
+        )
+
+    project = graphene.ID(source='project_id')
```
k9845, line 75: why project field here as well, since we have in fields?

apps/project/schema.py
```diff
@@ -0,0 +1,63 @@
...
+    # NOTE: This is a custom feature
+    # see: https://github.com/eamigo86/graphene-django-extras/compare/graphene-v2...the-deep:graphene-v2
+    @staticmethod
+    def get_custom_node(queryset, info, id):
+        try:
+            project = Project.get_for_gq(info.context.user).get(pk=id)
+            info.context.set_active_project(project)
+            return project
+        except Project.DoesNotExist:
+            return None
```
k9845: What does this method do?

thenav56: There is a comment there.

    # NOTE: This is a custom feature
    # see: https://github.com/eamigo86/graphene-django-extras/compare/graphene-v2...the-deep:graphene-v2

apps/lead/schema.py
thenav56: This is to force the field to be ID instead of ProjectType.

k9845 (approved):

Issue #872, opened 2021-07-21
DEEP Redesign: Add exclude filter for regions


subinasr:
**Insert Screenshot**

## Acceptance Criteria
* When filtered by projects, should return regions not in the project.

PR #873, opened 2021-07-21
Add filter to exclude regions in project

k9845:
Addresses #872 

## Changes
- Added exclude project regions filter in `/api/v1/regions/` api

@subinasr 

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/geo/filter_set.py
+++ b/apps/geo/filter_set.py
@@ -54,7 +54,7 @@ class RegionFilterSet(UserResourceFilterSet):
 
     def exclude_project_region_filter(self, qs, name, value):
         if value:
-            return qs.filter(~models.Q(project__in=value)).distinct()
+            return qs.exclude(project__in=value).distinct()
         return qs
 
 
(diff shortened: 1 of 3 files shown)


apps/geo/filter_set.py
```diff
@@ -16,3 +24,54 @@ def geo_area_label(self, queryset, name, value):
...
+                'filter_class': django_filters.CharFilter,
+                'extra': lambda f: {
+                    'lookup_expr': 'icontains',
+                },
+            },
+        }
+
+    def exclude_project_region_filter(self, qs, name, value):
+        if value:
+            return qs.filter(~models.Q(project__in=value)).distinct()
```
thenav56: This is fine but we can also use exclude here.
```suggestion
            return qs.exclude(project__in=value).distinct()
```

thenav56 (changes requested):

thenav56 (approved):

PR #874, opened 2021-07-21
Send label instead of title in discarded entry

AdityaKhatri:
## Changes

* Send label instead of title in discarded entry options

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/analysis/views.py
+++ b/apps/analysis/views.py
@@ -165,7 +165,7 @@ class DiscardedEntryOptionsView(views.APIView):
         options = [
             {
                 'key': tag.value,
-                'value': tag.name.title()
+                'value': tag.label,
             } for tag in DiscardedEntry.TagType
         ]
         return response.Response(options)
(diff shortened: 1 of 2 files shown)


apps/analysis/views.py
```diff
@@ -165,7 +165,7 @@ def get(self, request, version=None):
         options = [
             {
                 'key': tag.value,
-                'value': tag.name.title()
+                'value': tag.label,
```
thenav56, line 168: Please fix the test case as well.




Change `name.title()` to `label` here
https://github.com/the-deep/server/blob/10ba1dca74821210d5532ce4a6f7ecb4d5450881/apps/analysis/tests/test_apis.py#L1130-L1143

thenav56 (changes requested):

apps/analysis/views.py
AdityaKhatri: Okay sir!

thenav56 (approved):

k9845 (approved):

PR #886, opened 2021-07-27
Add members_count in user_groups/member_of api

k9845:
Addresses 
- New Project Flow: User Group issues(https://zube.io/deep/deep/c/3724)
## Changes
 - Added members_count in `api/v1/user-groups/member-of/?user={}` api

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/user_group/models.py
+++ b/apps/user_group/models.py
@@ -33,9 +33,9 @@ class UserGroup(UserResource):
 
     @staticmethod
     def get_for_member(user):
-        return UserGroup.objects.annotate(
-            members_count=models.Count('members', distinct=True)
-        ).filter(members=user).distinct()
+        return UserGroup.objects.filter(
+            members=user
+        ).distinct()
 
     @staticmethod
     def get_modifiable_for(user):
--- a/apps/user_group/models.py
+++ b/apps/user_group/models.py
@@ -33,9 +33,9 @@ class UserGroup(UserResource):
 
     @staticmethod
     def get_for_member(user):
-        return UserGroup.objects.filter(
-            members=user
-        ).distinct()
+        return UserGroup.objects.annotate(
+            members_count=models.Count('members', distinct=True)
+        ).filter(members=user).distinct()
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


apps/user_group/models.py
```diff
@@ -33,9 +33,9 @@ def get_for(user):
 
     @staticmethod
     def get_for_member(user):
-        return UserGroup.objects.filter(
-            members=user
-        ).distinct()
+        return UserGroup.objects.annotate(
+            members_count=models.Count('members', distinct=True)
+        ).filter(members=user).distinct()
```
thenav56: Let's annote in the views.

apps/user_group/models.py
```diff
@@ -33,9 +33,9 @@ def get_for(user):
 
     @staticmethod
     def get_for_member(user):
-        return UserGroup.objects.filter(
-            members=user
-        ).distinct()
+        return UserGroup.objects.annotate(
+            members_count=models.Count('members', distinct=True)
```
AdityaKhatri: Was members not used before?

AdityaKhatri (changes requested):

apps/user_group/models.py
thenav56: Not sure what you meant. But the filter is done by `.filter(members=user)`. @k9845 is only adding annotate here (members_count).

AdityaKhatri: I too am not sure what I mean, but just making sure this doesn't break the previous API usage.

thenav56: @k9845 Could you add a test if not present for the filter behavior?

thenav56 (approved):

apps/user_group/models.py
k9845: This is done @AdityaKhatri

AdityaKhatri (approved):

PR #887, opened 2021-07-27
Project: Graphql

thenav56:
Addresses (all related to graphql + new UI) (Base: https://github.com/the-deep/server/pull/836)

## Changes

- Change dockerfile (WARNING)
    - Use python buster image as base image.
    - Remove chrome/ploty (Tabular, thumbnail feature will not work) (~2.5GB size reduction)
        - Add serverless module for this feature in future.
    - Add poetry
        - Upgrade all the packages.
- Move requirements.txt to requirements.old.txt

- Add factory-boy.
- Refactor captcha logic. (Add Tests)

- All other issues related to graphql.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

## Note (WARNING)
Merging this to develop will have these major changes:
- We need to build a server after each poetry.lock/pyproject.toml change.
- We need to set up poetry/pyenv in local to add/remove packages. [Follow this](https://outline.togglecorp.com/share/37273873-8452-4c93-871a-76c169c6f6b0)
- Tabular Viz and Thumbnail generation will not work. (Due to docker image size reduction)
- Now project join request duplication is validated at the database level. TODO: Make sure production/staging doesn't have duplicate join requests.


apps/user/serializers.py
```diff
@@ -272,3 +278,169 @@ def save(self, **kwargs):
...
+    def is_captcha_required(cls, user=None, email=None):
+        _user = user or User.objects.filter(email=email).first()
+        return (
+            _user is not None and
+            _user.profile.login_attempts >= settings.MAX_LOGIN_ATTEMPTS_FOR_CAPTCHA
+        )
+
+    def validate_password(self, password):
+        # this will now only handle max-length in the login
+        CustomMaximumLengthValidator().validate(password=password)
```
bimal125, line 314: What about other auth validators?

bimal125 (commented):
There is one question in this PR @thenav56

apps/user/serializers.py
thenav56: This is a custom validator defined in the serializer. Auth validators are used by Django (this is used by DRF).

Not sure why this is not in the register. Do you remember @k9845?

k9845: Because many were facing weird validation errors on login for those whose password was created before this implementation. To fix this, we need to only validate max length on login and all others on the register.

thenav56: Okay, thanks.
To summarize, since we don't use Django auth validator for signup before, forcing them in login will make them impossible to log in. So, we only included max-length-validator in login.

bimal125 (dismissed):
Looks good for me

Issue #858, opened 2021-07-13
Support optimistic locking


tnagorra:
- There should be a global versionId which changes when any of the widgets change on an analytical framework.

thenav56:
@tnagorra @frozenhelium @AdityaKhatri 

Let's explore `modified_at` instead of `version_id`. `modified_at` is available and ready-to-be-used in graphql endpoints.

Issue #857, opened 2021-07-13
Query to get analytical framework


tnagorra:
* Metadata
* Primary tagging
* Secondary tagging
* Members

Issue #856, opened 2021-07-13
Mutation to save an analytical framework


tnagorra:
* Metadata
* Primary tagging
* Secondary tagging
* Members

Issue #853, opened 2021-07-13
The users should be able to only create N widgets and M sections.


tnagorra:
* The limits should be decided by the team.
* Excluding soft deleted entities
* where N=50, M=5

NOTE:
We cannot paginate the widgets and sections so we are limiting the total no. of items the users can create.

thenav56:
@Suraj-Shrestha @tnagorra @frozenhelium @AdityaKhatri 

Since we have secondary tagging, should the limit there be 50 as well? And 50 for every 5 sections?

thenav56:
~~We will have.~~
~~50 widgets max. (Af widgets limit)~~
~~5 sections max.~~

thenav56:
We will have this now.
    MAX_SECTIONS_ALLOWED = 5
    MAX_WIDGETS_ALLOWED_PER_SECTION = 10
    MAX_WIDGETS_ALLOWED_IN_SECONDARY_TAGGING = 50

So max widgets will be 5 * 10 + 50 = 100 (AF limit)

tnagorra:
Works for me!

PR #890, opened 2021-07-29
Feature/af graphql

thenav56:
Addresses
- #851 
- #857 
- #856 
- #895 
- #896
- #853
- #858

## Changes

* Add Queries for AF.
* Add Mutations for AF.
* Add lastActiveProject in me Query.
* Add accessibleFeatures in me Query.
* Add section.
* Add widgets count limit in AF.
* Add test to make sure modifiedAt is updated [for AF] after mutation.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations


apps/analysis_framework/mutation.py
```diff
@@ -1,45 +1,48 @@
...
         data = AnalysisFrameworkInputType(required=True)
-        id = graphene.ID(required=True)
 
+    # output fields
     result = graphene.Field(AnalysisFrameworkType)
     # class vars
     serializer_class = AnalysisFrameworkSerializer
     model = AnalysisFramework
-    permissions = []  # TODO:
+    permissions = []
```
k9845: Should not we handle permission for creating?

k9845 (approved):

bimal125 (approved):
Looks good.

apps/analysis_framework/mutation.py
thenav56: AnalysisFramework is a root entity like a project, so creating one doesn't require any permission.

apps/analysis_framework/serializers.py
```diff
@@ -289,3 +282,183 @@ def update(self, instance, validated_data):
 
     def get_is_admin(self, analysis_framework):
         return analysis_framework.can_modify(self.context['request'].user)
+
+
+# ------------------ Graphql seriazliers -----------------------------------
+def validate_items_limit(items, limit, error_message='Only %d items are allowed. Provided: %d'):
+    if items:
+        count = len(items)
+        print(count, limit)
```
k9845: No `print`

apps/analysis_framework/serializers.py
```diff
@@ -289,3 +282,183 @@ def update(self, instance, validated_data):
...
+        if count > limit:
+            raise serializers.ValidationError(error_message % (limit, count))
+
+
+class AfWidgetLimit():
+    MAX_SECTIONS_ALLOWED = 5
+    MAX_WIDGETS_ALLOWED_PER_SECTION = 10
+    MAX_WIDGETS_ALLOWED_IN_SECONDARY_TAGGING = 50
```
k9845, line 298: Can we define this in `settings` or here is better?

apps/analysis_framework/serializers.py
thenav56: Good catch.

apps/analysis_framework/serializers.py
thenav56: This is only for serializer validation. This is only for bad actor scenarios.

deep/serializers.py
```diff
@@ -137,22 +132,29 @@ def get_fields(self, *args, **kwargs):
...
+class TempClientIdMixin(serializers.ModelSerializer):
     """
     ClientId for serializer level only, not storing to the database.
     """
     client_id = serializers.CharField(required=False)
 
+    def _get_temp_client_id(self, validated_data):
+        try:
+            self.Meta.model._meta.get_field('client_id')
+            return None
```
bimal125, line 144: This always returns None please confirm .

thenav56: Yes, we don't want to remove **client_id** from validated_data if the model has a **client_id** field. so we just return None here. But if the model doesn't have the field then it will return (with pop -> remove from **valdiated_data** -> not passed to model.create()/model.update()).

I will make sure to add some comments here in the next PR.

thenav56: I will send this on the next PR.

Issue #919, opened 2021-08-02
Create mutation to cancel project join request


AdityaKhatri:
### New Addition
* Add mutation to cancel project join request

Issue #894, opened 2021-08-02
Create mutation for project join request


AdityaKhatri:
### New changes
* Create mutation for project join request
* Should support description (min length 50)
* Should create a notification and email alert like in the previous rest request

PR #921, opened 2021-08-18
Mutation for project join and cancel

k9845:
Addresses 
- https://github.com/the-deep/server/issues/894
- https://github.com/the-deep/server/issues/919

## Changes
- Added mutation to join and cancel project join

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/notification/receivers/project_membership.py
+++ b/apps/notification/receivers/project_membership.py
@@ -7,14 +7,13 @@ from project.models import (
     ProjectMembership,
     ProjectRole,
 )
-from project.serializers import ProjectJoinRequestSerializer
+from project.serializers import ProjectJoinGqSerializer
 
 
 @receiver(post_save, sender=ProjectJoinRequest)
 def create_notification(sender, instance, created, **kwargs):
     admins = instance.project.get_admins()
-    data = ProjectJoinRequestSerializer(instance).data
-
+    data = ProjectJoinGqSerializer(instance).data
     if (created):
         for admin in admins:
             Notification.objects.create(
@@ -62,7 +61,7 @@ def update_notification_for_join_request(sender, instance, **kwargs):
         notification.save()
 
     admins = instance.project.get_admins()
-    data = ProjectJoinRequestSerializer(instance).data
+    data = ProjectJoinGqSerializer(instance).data
     data['status'] = 'aborted'
     for admin in admins:
         Notification.objects.create(
@@ -102,7 +101,7 @@ def remove_notifications_for_former_project_admin(
                     receiver=instance.member,
                     notification_type=Notification.PROJECT_JOIN_REQUEST,
                     project=instance.project,
-                    data=ProjectJoinRequestSerializer(
+                    data=ProjectJoinGqSerializer(
                         old_project_join_request).data,
                 )
     except ProjectMembership.DoesNotExist:
@@ -124,6 +123,6 @@ def create_notifications_for_new_project_admin(
                     receiver=instance.member,
                     notification_type=Notification.PROJECT_JOIN_REQUEST,
                     project=instance.project,
-                    data=ProjectJoinRequestSerializer(
+                    data=ProjectJoinGqSerializer(
                         old_project_join_request).data,
                 )
--- a/apps/notification/receivers/project_membership.py
+++ b/apps/notification/receivers/project_membership.py
@@ -7,13 +7,13 @@ from project.models import (
     ProjectMembership,
     ProjectRole,
 )
-from project.serializers import ProjectJoinGqSerializer
+from project.serializers import ProjectJoinRequestSerializer
 
 
 @receiver(post_save, sender=ProjectJoinRequest)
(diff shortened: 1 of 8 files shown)


apps/notification/receivers/project_membership.py
```diff
@@ -7,14 +7,13 @@
...
-from project.serializers import ProjectJoinRequestSerializer
+from project.serializers import ProjectJoinGqSerializer
 
 
 @receiver(post_save, sender=ProjectJoinRequest)
 def create_notification(sender, instance, created, **kwargs):
     admins = instance.project.get_admins()
-    data = ProjectJoinRequestSerializer(instance).data
-
+    data = ProjectJoinGqSerializer(instance).data
```
thenav56: Let's not change this.

For Graphql we only need writable fields. So remove read-only fields from `ProjectJoinGqSerializer` and use `ProjectJoinRequestSerializer` here.

apps/project/mutation.py
```diff
@@ -1,9 +1,68 @@
...
+    ok = graphene.Boolean()
+    result = graphene.Field(ProjectJoinRequestType)
+
+    @staticmethod
+    def mutate(root, info, data):
+        serializer = ProjectJoinGqSerializer(data=data, context={'request': info.context.request})
+        if errors := mutation_is_not_valid(serializer):
+            return CreateProjectJoin(errors=errors, ok=False)
+        instance = serializer.save()
+        return CreateProjectJoin(result=instance, errors=None, ok=True)
```
thenav56, line 85: Let's take this opportunity to create a generic mutation without permission (we have several with permission). Check `utils/graphene/mutation.py`.

apps/project/mutation.py
```diff
@@ -1,9 +1,68 @@
...
+            return DeleteProjectJoin(errors=[
+                dict(field='nonFieldErrors', messages=gettext('ProjectJoinRequest does not exist.'))
+            ], ok=False)
+        instance.delete()
+        return DeleteProjectJoin(result=instance, errors=None, ok=True)
+
+
+class CreateProjectJoin(graphene.Mutation):
+    class Arguments:
+        data = ProjectJoinRequestInputType(required=True)
```
thenav56, line 73: We only need id here as an argument.

thenav56 (changes requested):

apps/project/mutation.py
```diff
@@ -1,9 +1,68 @@
...
+
+ProjectJoinCancelInputType = generate_input_type_for_serializer(
+    'ProjectJoinCancelInputType',
+    serializer_class=ProjectJoinCancelGqSerializer,
+)
+
+
+class DeleteProjectJoin(graphene.Mutation):
+    class Arguments:
+        data = ProjectJoinCancelInputType(required=True)
```
thenav56: We don't need any arguments here. Just delete a ProjectRequest for the current user in the current project if exists.

apps/project/tests/test_schemas.py
```diff
@@ -142,3 +152,223 @@ def test_project_query_filter(self):
...
+    def test_join_private_project(self):
+        user = UserFactory.create()
+        project = ProjectFactory.create(is_private=True)
+        reason = "\
+          You gotta be crazy, you gotta have a real need\
+          You gotta sleep on your toes, and when you're on the street\
+          You gotta be able to pick out the easy meat with your eyes closed\
+          And then moving in silently, down wind and out of sight\
+          You gotta strike when the moment is right without thinking\
+        "
```
thenav56: We can also use fuzzy to generate this automatically.
```python
FuzzyText(length=<length>).fuzz()
```

apps/project/serializers.py
```diff
@@ -30,6 +36,8 @@
 from .permissions import PROJECT_PERMISSIONS
 from .activity import project_activity_log
 
+MIN_LENGTH = 50
```
thenav56: Let's define this inside the serializer.

apps/project/serializers.py
```diff
@@ -490,3 +498,106 @@ class ProjectRecentActivitySerializer(serializers.Serializer):
...
+        role = validated_data.pop('role', None)
+        if instance.status in ['accepted', 'rejected']:
+            raise serializers.ValidationError(
+                'This request has already been {}'.format(instance.status)
+            )
+        if validated_data['status'] == 'accepted':
+            ProjectAcceptRejectSerializer._accept_request(self.context['request'].user, instance, role)
+        elif validated_data['status'] == 'rejected':
+            ProjectAcceptRejectSerializer._reject_request(self.context['request'].user, instance)
+        return super().update(instance, validated_data)
```
thenav56: Looks like we can merge this two serializer now. Let's use partial=True, that way create validator isn't triggered in update (if they are provided throw an error if different values are provided)

thenav56 (changes requested):

PR #928, opened 2021-08-25
Add authoring_organization_type in excel export

k9845:
Addresses 

## Changes
- Added `authoring_oragnizations_type` in excel export for entry

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/apps/export/entries/excel_exporter.py
+++ b/apps/export/entries/excel_exporter.py
@@ -57,6 +57,7 @@ class ExcelExporter:
             'Lead Id',
             'Lead Title',
             'Lead URL',
+            'Authoring Organizations Type',
             'Author',
             'Source',
             'Lead Priority',
@@ -402,7 +403,6 @@ class ExcelExporter:
 
             lead = entry.lead
             assignee = entry.lead.get_assignee()
-
             author = lead.get_authors_display()
             source = lead.get_source_display()
             published_on = (lead.published_on and lead.published_on.strftime(EXPORT_DATE_FORMAT)) or ''
@@ -420,6 +420,7 @@ class ExcelExporter:
                 f'{lead.id}',
                 lead.title,
                 lead.url or (lead.attachment and lead.attachment.get_file_url()),
+                lead.get_authoring_organizations_type_display(),
                 lead.get_authors_display(),
                 lead.get_source_display(),
                 lead.get_priority_display(),
--- a/apps/export/tasks/tasks_entries.py
+++ b/apps/export/tasks/tasks_entries.py
@@ -55,9 +55,10 @@ def export_entries(export):
     queryset = Entry.get_exportable_queryset(queryset)\
         .prefetch_related(
             'lead__authors',
+            'lead__authors__organization_type',
+            # Also organization parents
             'lead__authors__parent',
             'lead__authors__parent__organization_type',
-            'lead__authors__organization_type'
         )
 
     search = filters.get('search')
--- a/apps/export/tasks/tasks_entries.py
+++ b/apps/export/tasks/tasks_entries.py
@@ -52,7 +52,13 @@ def export_entries(export):
     queryset = queryset.prefetch_related(
         'entrygrouplabel_set'
     )
-    queryset = Entry.get_exportable_queryset(queryset)
+    queryset = Entry.get_exportable_queryset(queryset)\
+        .prefetch_related(
+            'lead__authors',
+            'lead__authors__parent',
+            'lead__authors__parent__organization_type',
+            'lead__authors__organization_type'
+        )
 
     search = filters.get('search')
     if search:
--- a/apps/lead/models.py
+++ b/apps/lead/models.py
@@ -283,7 +283,7 @@ class Lead(UserResource, ProjectEntityMixin):
         authors = self.authors.all()
         if authors:
             return ','.join(set([
-                author.get_organization_type_display() for author in authors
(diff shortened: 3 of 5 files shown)


apps/organization/models.py
```diff
@@ -69,3 +69,8 @@ def __str__(self):
         return f'{self.pk} : ({self.short_name}) {self.title} ' + (
             '(MERGED)' if self.parent else ''
         )
+
+    def get_organization_type_display(self):
+        if self.parent:
+            return self.parent.organization_type.title
+        return self.organization_type.title
```
thenav56: Let's use `self.data` for parent/self reference.
```suggestion
    def get_organization_type_display(self):
        return self.data.organization_type.title
```

apps/export/entries/excel_exporter.py
```diff
@@ -420,6 +420,7 @@ def add_entries(self, entries):
                 f'{lead.id}',
                 lead.title,
                 lead.url or (lead.attachment and lead.attachment.get_file_url()),
+                lead.get_authoring_organizations_type_display(),
```
thenav56, line 423: Also, add this `authors, authors parent, authors parent organization type, authors organization type` prefetch in `Entry.get_exportable_queryset)` as well.

apps/lead/models.py
```diff
@@ -277,6 +277,13 @@ def get_authors_display(self):
             return self.author and self.author.data.title
         return self.author_raw
 
+    def get_authoring_organizations_type_display(self):
+        if self.authors.exists():
+            return ','.join([author.get_organization_type_display() for author in self.authors.all()])
```
thenav56: We need to use `set` here, multiple authors can have the same organization type.

```suggestion
            return ','.join(set([
                author.get_organization_type_display() for author in self.authors.all()
            ]))
```

thenav56 (changes requested):

apps/export/tasks/tasks_entries.py
```diff
@@ -52,7 +52,13 @@ def export_entries(export):
     queryset = queryset.prefetch_related(
         'entrygrouplabel_set'
     )
-    queryset = Entry.get_exportable_queryset(queryset)
+    queryset = Entry.get_exportable_queryset(queryset)\
+        .prefetch_related(
+            'lead__authors',
+            'lead__authors__parent',
+            'lead__authors__parent__organization_type',
+            'lead__authors__organization_type'
```
thenav56:
```suggestion
            'lead__authors',
            'lead__authors__organization_type',
            # Also organization parents            
            'lead__authors__parent',
            'lead__authors__parent__organization_type',
```

apps/lead/models.py
```diff
@@ -277,6 +277,15 @@ def get_authors_display(self):
             return self.author and self.author.data.title
         return self.author_raw
 
+    def get_authoring_organizations_type_display(self):
+        if self.authors.exists():
+            return ','.join(set([
+                author.get_organization_type_display() for author in self.authors.all()
+            ]))
```
thenav56: Let's do it like this. Since we are prefetching, doing `exists` will send another request to the database.
Also, update the `get_authors_display`.

```suggestion
        authors = self.authors.all()
        if authors:
            return ','.join(set([
                author.get_organization_type_display() for author in authors
            ]))
```

thenav56 (changes requested):

apps/lead/models.py
```diff
@@ -277,6 +277,15 @@ def get_authors_display(self):
             return self.author and self.author.data.title
         return self.author_raw
 
+    def get_authoring_organizations_type_display(self):
```
thenav56, line 282: Also, let's add unit tests for these two (authors and organization type) functions as well.

apps/lead/models.py
```diff
@@ -269,14 +269,26 @@ def get_source_display(self):
...
+        authors = self.authors.all()
+        if authors:
+            return ','.join(set([
+                author.get_organization_type_display() for author in authors
+            ]))
+        elif self.author:
+            return self.author.organization_type and self.author.data.organization_type.title
```
thenav56:
```suggestion
            return self.author.data.organization_type and self.author.data.organization_type.title
```

apps/lead/tests/test_apis.py
```diff
@@ -1562,6 +1562,50 @@ def test_lead_group_get(self):
...
+        organization1 = self.create(Organization, organization_type=organization_type1)
+        organization2 = self.create(Organization, organization_type=organization_type2)
+        organization3 = self.create(Organization, organization_type=organization_type3)
+        organization4 = self.create(Organization, parent=organization1)
+
+        # create lead
+        lead1 = self.create_lead(project=project, authors=[organization3, organization2])
+        lead2 = self.create_lead(project=project, authors=[organization4])
+        # just provide author for this
+        lead3 = self.create_lead(project=project, author=organization1)
```
thenav56, line 1583: Add another lead with organization1 and organization4 (with the same organization type) as well.

thenav56 (changes requested):

apps/export/tasks/tasks_entries.py
thenav56: Please do the ordering as well.

thenav56 (dismissed):

thenav56 (approved):

Issue #968, opened 2021-09-21
If new entry is added to `tagged` lead, change the status to `in progress`


AdityaKhatri:
### New Addition
If new entry is added or old entry is updated to `tagged` lead, change the status to `in progress`

Issue #966, opened 2021-09-21
Change lead statuses to `not tagged`, `in progress` and `tagged`


AdityaKhatri:
### New Addition
* Add `not tagged` status
* Change `validated` to `tagged`
* Change `processed` to `tagged`

PR #977, opened 2021-09-23
Auto change lead status.

thenav56:
Addresses
- #966
- #968

## Changes

- Add receiver to change lead status to IN_PROGRESS
- Move mutation test to separate file for lead.
- Change Lead Status Enum values.

@AdityaKhatri Lead Status Enum is changed.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations


k9845 (dismissed):
Shouldn't we change the older lead statuses and map them with the new statuses?

thenav56:
> Shouldn't we change the older lead statuses and map them with the new statuses?

We are still not sure if this is the final status. For now, just changing the enum labels which should be enough for now. There is also a TODO.

k9845:
> We are still not sure if this is the final status. For now, just changing the enum labels which should be enough for now. There is also a TODO.

Sounds cool.

bimal125 (dismissed):
looks good for me

k9845 (approved):

Issue #854, opened 2021-07-13
Add validation for each widget when user saves the analytical framework/entry


tnagorra:
- Each fields in the widget should be validated by the server.
- The validation depends on the type of widget created.
- The server should send proper error message to the client when an error occurs.

PR #994, opened 2021-10-08
Fix lead filter not working with entry data.

thenav56:
Addresses Lead filter not working with entry filter data

## Changes
- Fix filter issues.
- Add appropriate test cases.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/entry/tests/test_schemas.py
+++ b/apps/entry/tests/test_schemas.py
@@ -7,6 +7,8 @@ from analysis_framework.models import Widget
 
 from utils.common import ONE_DAY
 from utils.graphene.tests import GraphQLTestCase
+from lead.filter_set import LeadFilterSet
+
 from user.factories import UserFactory
 from geo.factories import RegionFactory, AdminLevelFactory, GeoAreaFactory
 from project.factories import ProjectFactory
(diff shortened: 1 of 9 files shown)


apps/entry/tests/test_schemas.py
```diff
@@ -548,7 +568,7 @@ def test(self):
         EntryAttributeFactory.create(
             entry=entry3_2,
             widget=self.widget_date_range,
-            data={'value': {'startDate': '2020-01-15', 'endDate': '2020-01-10'}},
+            data={'value': {'startDate': '2020-01-15', 'endDate': '2020-01-25'}},
```
k9845, line 571: may be write validation for this?

k9845 (approved):

apps/entry/tests/test_schemas.py
thenav56: Write validation for entry attributes? We have a separate ticket #854 for that.

k9845: yes.

bimal125 (approved):

Issue #1018, opened 2021-11-12
DEEP Stats: Make query for most used Analysis Framework


subinasr:
![image.png](https://zube.io/files/deep/b941078466dc2877e0abb0c34956c019-image.png)

## Acceptance Criteria
- Instead of 'most active Organizations' as shown in screenshot, the query should give top 5 most used Analysis Frameworks
- Project count for each analysis framework should also be in the stats query

### Attachments
- [Figma Link](https://www.figma.com/proto/HaQd4COVlpSMoqSfgL5ZDx/DEEP---UI_Library-%2B-Screens?node-id=10324%3A80896&scaling=min-zoom&hide-ui=1)

PR #1021, opened 2021-11-16
Add recent_analysis_framework query

k9845:
Addresses 
- https://github.com/the-deep/server/issues/1018

## Changes
- Add query for recent_framework based on used in project

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/project/schema.py
+++ b/apps/project/schema.py
@@ -106,16 +106,11 @@ class ProjectExploreStatType(graphene.ObjectType):
             type('ExploreProjectStatTopActiveFrameworksType', (graphene.ObjectType,), {
                 'analysis_framework_id': graphene.Field(graphene.NonNull(graphene.ID)),
                 'analysis_framework_title': graphene.String(),
-                'project_count': graphene.Int(),
-                'source_count': graphene.Int()
+                'project_count': graphene.NonNull(graphene.Int),
+                'source_count': graphene.NonNull(graphene.Int)
             })
         )
     )
-
-    @staticmethod
-    def resolve_top_active_frameworks(root, info, **kwargs):
-        from analysis_framework.models import AnalysisFramework
-        return AnalysisFramework.get_recent_frameworks(root, info.context.user)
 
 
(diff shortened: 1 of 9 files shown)


apps/project/schema.py
```diff
@@ -98,6 +98,21 @@ class ProjectExploreStatType(graphene.ObjectType):
...
+                'project_count': graphene.Int(),
+                'source_count': graphene.Int()
+            })
+        )
+    )
+
+    @staticmethod
+    def resolve_top_active_frameworks(root, info, **kwargs):
+        from analysis_framework.models import AnalysisFramework
+        return AnalysisFramework.get_recent_frameworks(root, info.context.user)
```
thenav56: - Let define this in project dataloader `ProjectExploreStatsLoader`, It handles cache.

- We can also use **only** or values here.
  https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.query.QuerySet.only

apps/project/schema.py
```diff
@@ -98,6 +98,21 @@ class ProjectExploreStatType(graphene.ObjectType):
             })
         )
     )
+    top_active_frameworks = graphene.List(
+        graphene.NonNull(
+            type('ExploreProjectStatTopActiveFrameworksType', (graphene.ObjectType,), {
+                'analysis_framework_id': graphene.Field(graphene.NonNull(graphene.ID)),
+                'analysis_framework_title': graphene.String(),
+                'project_count': graphene.Int(),
+                'source_count': graphene.Int()
```
thenav56: We can use NonNull when we use Coalesce.

thenav56 (changes requested):

thenav56 (approved):
All good. Two things to do.
- Move AnalysisFramework import to top in project dataloaders.py
- Re-generate schema.graphql

thenav56 (approved):

PR #1078, opened 2022-04-11
Add default organization order

thenav56:
## Changes

- Add default organization order
- Fix email lower in login

- [ ] n+1 queries
- [ ] flake8 issues
- [ ] `print`
- [ ] typos
- [ ] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/apps/user/serializers.py
+++ b/apps/user/serializers.py
@@ -209,14 +209,13 @@ class PasswordResetSerializer(RemoveNullFieldsMixin,
     email = serializers.EmailField(required=True)
 
     def get_user(self, email):
-        users = User.objects.filter(email=email)
+        users = User.objects.filter(
+            models.Q(email=email.lower()) |
+            models.Q(email=email)
+        )
         if not users.exists():
             raise UserNotFoundError
         return users.first()
-
-    def validate_email(self, email):
-        self.get_user(email)
-        return email
 
     def validate_hcaptcha_response(self, captcha):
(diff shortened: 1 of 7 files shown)


apps/user/serializers.py
```diff
@@ -320,11 +319,18 @@ def _set_user_login_attempts(user, login_attempts):
             user.profile.login_attempts = login_attempts
             user.profile.save(update_fields=['login_attempts'])
 
+        email = data['email']
         # NOTE: authenticate only works for active users
         # NOTE: username should be equal to email
-        authenticate_user = authenticate(username=data['email'], password=data['password'])
+        authenticate_user = authenticate(username=email.lower(), password=data['password'])
```
k9845, line 325: Just a question why are we converting email to `lower`?

k9845 (approved):

apps/user/serializers.py
thenav56: just to make sure the same email without different cases is not allowed in the deep.

thenav56:
@kumud02 
Test case for this.
- Organization: @AdityaKhatri will fill this.
- User email lowercase:
  - We should be able to add users with different cases from the Django admin panel. (For adding bad emails to the system)
  - Users with different cases should be able to login into the deep with separate credentials. (To support legacy users)
  - User shouldn't be able to register with the same email (but with different cases).

PR #1119, opened 2022-06-03
Add api for recent activity in graphql

Rup-Narayan-Rajbanshi:
## Changes

- Add api for recent activity in graphql

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [x] translations

### DIFF
--- a/apps/project/models.py
+++ b/apps/project/models.py
@@ -27,6 +27,12 @@ from django.utils import timezone
 from datetime import timedelta
 
 
+class RecentActivityType(models.TextChoices):
+    LEAD = 'lead', 'Source'
+    ENTRY = 'entry', 'Entry'
+    ENTRY_COMMENT = 'entry-comment', 'Entry Comment'
+
+
 class Project(UserResource):
     """
     Project model
@@ -276,7 +282,6 @@ class Project(UserResource):
             _get_activities,
             60 * 5,  # 5min
         )
-
         return [
             {
                 field: item[index]
--- a/apps/project/schema.py
+++ b/apps/project/schema.py
@@ -25,6 +25,7 @@ from deep.serializers import URLCachedFileField
 from user_resource.schema import UserResourceMixin
 
 from user.models import User
+from user.schema import UserType
 from lead.schema import Query as LeadQuery
 from entry.schema import Query as EntryQuery
 from export.schema import Query as ExportQuery
@@ -49,6 +50,7 @@ from .models import (
     ProjectJoinRequest,
     ProjectOrganization,
     ProjectStats,
+    RecentActivityType as ActivityTypes,
 )
 from .enums import (
     ProjectPermissionEnum,
@@ -57,6 +59,7 @@ from .enums import (
     ProjectJoinRequestStatusEnum,
     ProjectOrganizationTypeEnum,
     ProjectMembershipBadgeTypeEnum,
+    RecentActivityTypeEnum,
 )
 
 from .filter_set import (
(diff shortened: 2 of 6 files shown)


apps/project/schema.py
```diff
@@ -116,6 +116,21 @@ def get_top_entity_contributor(project, Entity):
     ]
 
 
+class RecentActivityType(graphene.ObjectType):
+    id = graphene.Int()
+    created_at = graphene.DateTime()
+    project = graphene.Int()
```
thenav56:
```suggestion
    project = graphene.ID(required=True)
```

apps/project/schema.py
```diff
@@ -116,6 +116,21 @@ def get_top_entity_contributor(project, Entity):
     ]
 
 
+class RecentActivityType(graphene.ObjectType):
+    id = graphene.Int()
```
thenav56:
```suggestion
    id = graphene.ID(required=True)
```

apps/project/schema.py
```diff
@@ -116,6 +116,21 @@ def get_top_entity_contributor(project, Entity):
...
+    created_at = graphene.DateTime()
+    project = graphene.Int()
+    project_display_name = graphene.String()
+    created_by = graphene.Int()
```
thenav56:
```suggestion
    created_by = graphene.ID()
```

apps/project/schema.py
```diff
@@ -116,6 +116,21 @@ def get_top_entity_contributor(project, Entity):
...
+    created_by_display_picture = graphene.String()
+    type = graphene.String()
```
thenav56: Let's use enum instead of String.

apps/project/schema.py
thenav56: Maybe we can just use usertype here.

thenav56 (changes requested):

apps/project/models.py
```diff
@@ -27,6 +27,12 @@
 from datetime import timedelta
 
 
+class RecentActivityType(models.TextChoices):
+    LEAD = 'lead', 'lead'
+    ENTRY = 'entry', 'entry'
+    ENTRY_COMMENT = 'entry-comment', 'entry-comment'
```
thenav56:
```suggestion
    LEAD = 'lead', 'Source'
    ENTRY = 'entry', 'Entry'
    ENTRY_COMMENT = 'entry-comment', 'Entry Comment'
```

apps/project/schema.py
```diff
@@ -295,6 +297,26 @@ def resolve_regions(root, info, **kwargs):
...
+    def resolve_created_by(root, info, **kwargs):
+        id = int(root['created_by'])
+        return info.context.dl.project.users.load(id)
+
+    def resolve_project(root, info, **kwargs):
+        id = int(root['project'])
+        return info.context.dl.project.projects.load(id)
+
+    def resolve_type_display(root, info, **kwargs):
+        return root['type']
```
thenav56: Can you confirm this provides label instead of value?

thenav56 (changes requested):

apps/project/schema.py
Rup-Narayan-Rajbanshi: @thenav56 , here type is not generated from Choice fieds insead it is generate as models.Value('entry-comment', output_field=models.CharField()), so it return text string.

thenav56 (approved):

AdityaKhatri:
Closed because it is included in this: https://github.com/the-deep/server/pull/1128/commits

PR #1174, opened 2022-11-12
Show all leads in the excel bibliography.

thenav56:
## Changes

* Show all leads in the excel bibliography.
* Add entries count.

- [ ] n+1 queries
- [ ] flake8 issues
- [ ] `print`
- [ ] typos
- [ ] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/apps/export/entries/excel_exporter.py
+++ b/apps/export/entries/excel_exporter.py
@@ -1,5 +1,6 @@
 import logging
 from django.core.files.base import ContentFile
+from django.db import models
 
 from deep.permalinks import Permalink
 from utils.common import (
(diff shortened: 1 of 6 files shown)


apps/export/entries/excel_exporter.py
```diff
@@ -36,7 +37,8 @@ class ColumnsData:
             ] if self.modified_excerpt_exists else ['Excerpt'],
         }
 
-    def __init__(self, export_object, entries, columns=None, decoupled=True, project_id=None, is_preview=False):
+    def __init__(self, export_object, entries, project, columns=None, decoupled=True, is_preview=False):
```
bewakes, line 40: Since project object is not used inside, why was project object passed instead of project_id? Just out of curiosity.

bewakes (approved):

PR #1195, opened 2022-12-16
Feature/lead deduplication

### LINKED ISSUE #1267: Source Deduplication

bewakes:
Resolves: #1267 
Depends:

## Changes

* Callback api for deduplication request from nlp server
* Update lead model to add duplicate leads
* Update lead schema to include duplicate leads

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations


apps/lead/serializers.py
```diff
@@ -717,3 +717,42 @@ def validate(self, data):
...
+            raise serializers.ValidationError("Invalid data. Provide lead_id and duplicate_lead_ids")
+
+        project_leads = self.duplicate_leads.filter(project=self.lead.project)
+        if project_leads.count() != len(data["duplicate_lead_ids"]):
+            raise serializers.ValidationError("Leads belong to different project")
+        return data
+
+    def create(self, validated_data):
+        if self.lead is None or self.duplicate_leads is None:
+            raise serializers.ValidationError("Invalid data. Provide lead_id and duplicate_lead_ids")
```
thenav56: Not required.

apps/lead/schema.py
```diff
@@ -395,7 +395,20 @@ def resolve_filtered_entries_count(root, info, **kwargs):
...
+
+    class Meta:
+        model = Lead
+        only_fields = (
+            'id', 'title', 'is_assessment_lead', 'lead_group', 'assignee', 'published_on',
+            'text', 'url', 'attachment',
+            'client_id',
+        )
+
+
```
thenav56: Let's do this
- Add duplicates count in LeadType
- Add duplicates count 0 or >0 filter in LeadType Listing
- Add pagination query for DuplicateLeadsType with lead_id as an argument. In return use LeadType Listing

thenav56 (changes requested):
Looks good. Small changes.

apps/lead/filter_set.py
```diff
@@ -325,6 +326,14 @@ class LeadGQFilterSet(UserResourceGqlFilterSet):
     emm_keywords = django_filters.CharFilter(method='emm_keywords_filter')
     emm_risk_factors = django_filters.CharFilter(method='emm_risk_factors_filter')
 
+    # duplicates
+    duplicate_leads_count = django_filters.NumberFilter()
+    duplicate_leads_count__gte = django_filters.NumberFilter(
+        field_name='duplicate_leads', lookup_expr='gte')
+    duplicate_leads_count__lte = django_filters.NumberFilter(
+        field_name='duplicate_leads', lookup_expr='lte')
```
thenav56: I think has_duplicates should be enough for now.
cc: @AdityaKhatri

thenav56 (changes requested):
LGTM, just one change request

apps/lead/serializers.py
```diff
@@ -717,3 +720,42 @@ def validate(self, data):
         data['project'] = self.project
         data['user'] = self.current_user
         return data
+
+
+class DeduplicationCallbackSerializer(serializers.Serializer):
+    lead_id = serializers.IntegerField()
```
k9845: Shouldn't this be char?

k9845 (approved):

apps/lead/serializers.py
thenav56: Nice catch. Yes, we need to use `client_id` (char) instead of lead_id. As we use callback without authentication, we use client_id to validate the request. client_id is generated using lead_id and a signature using Django-secret. You can check out ExtractCallbackSerializer for more info.

bewakes: Done!

apps/lead/filter_set.py
bewakes: Done!

apps/lead/filter_set.py
```diff
@@ -527,6 +532,27 @@ def _entry_subquery(entry_qs: models.QuerySet):
...
+            return qs.filter(
+                models.Q(duplicate_leads__isnull=False) |
+                models.Q(duplicate_of__isnull=False),
+            )
+        elif val is False:
+            return qs.filter(
+                duplicate_leads=None,
+                duplicate_of=None,
+            )
+        return qs
```
thenav56: We can use the annotated count here?

apps/lead/serializers.py
```diff
@@ -717,3 +720,51 @@ def validate(self, data):
...
+        if self.lead is None or self.duplicate_leads is None:
+            raise serializers.ValidationError("Invalid data. Provide lead_id and duplicate_lead_ids")
+
+        # Also validate if lead_id corresponds to the client id
+        try:
+            lead = LeadExtraction.get_lead_from_client_id(data['client_id'])
+        except LeadExtraction.Exception.InvalidTokenValue:
+            raise serializers.ValidationError("Invalid token")
+        if lead.id != data['lead_id']:
+            raise serializers.ValidationError("Inconsistent lead_id and client_id")
```
k9845: Why error message has `client_id`?

thenav56 (changes requested):
LGTM

k9845 (approved):

apps/lead/serializers.py
bewakes: Because it is checking whehter the lead  id is associated with corresponding client_id or not. Such checks are not required in other callbacks but it's done here.

apps/lead/filter_set.py
bewakes: That's a good point. Will test and push.

thenav56 (approved):

k9845 (approved):

apps/deduplication/models.py
```diff
@@ -0,0 +1,140 @@
...
+    glove_840b_300d        0.525          0.7385           9.7624               0.00705
+
+    datasketch_lsh_perm256 0.55           0.7981           52.82732             6.4
+    datasketch_lsh_perm128 0.55           0.7816           37.68474             4.81983
+    """
+    NUM_PERM = 256
+
+    class IndexStatus(models.TextChoices):
+        CREATING = "creating", "Creating"
+        CREATED = "created", "Created"
```
thenav56: Let's use Integer if we can.

apps/deduplication/models.py
```diff
@@ -0,0 +1,140 @@
...
+
+    name = models.CharField(max_length=256)
+    status = models.CharField(
+        max_length=20,
+        choices=IndexStatus.choices,
+        default=IndexStatus.CREATING,
+    )
+    project = models.OneToOneField(
+        Project,
+        null=True,
```
thenav56: Is there a reason for using `null=True`?

apps/deduplication/models.py
```diff
@@ -0,0 +1,140 @@
...
+    text_extract = models.TextField()
+    has_errored = models.BooleanField(default=False)
+    error = models.TextField(null=True, blank=True)
+    result = models.JSONField(default=dict)
+
+    class Meta:
+        unique_together = ("project_id", "lead_id")
+
+    def __str__(self):
+        return f"{self.status}: Project({self.project_id}) Lead({self.lead_id})"
```
thenav56: This is not required now, right?

apps/deduplication/models.py
```diff
@@ -0,0 +1,140 @@
...
+        default=IndexStatus.CREATING,
+    )
+    project = models.OneToOneField(
+        Project,
+        null=True,
+        on_delete=models.CASCADE,
+        unique=True,
+    )
+    pickle_version = models.CharField(max_length=10, null=True)
+    index_pickle = models.BinaryField(null=True)
```
thenav56, line 48: Can you share a bit more about this? How does index_pickle size increases?

apps/lead/views.py
```diff
@@ -174,6 +175,13 @@ def get_queryset(self):
             ).filter(similarity__gt=0.3).order_by('-similarity')
         return leads
 
+    def destroy(self, *args, **kwargs):
+        lead = self.get_object()
+        result = super().destroy(*args, **kwargs)
+        # remove from index after the deletion is successful
+        remove_lead_from_index.delay(lead.id)
+        return result
+
```
thenav56: We will need to use signals as DRF views are not used anymore.

apps/deduplication/tasks/indexing.py
```diff
@@ -0,0 +1,148 @@
...
+                index_obj.save()
+    except Exception:
+        logger.warning(
+            f"Error creating index for project {project.title}({project.id})",
+            exc_info=True
+        )
+        import traceback
+
+        index_obj.has_errored = True
+        index_obj.error = traceback.format_exc()
```
thenav56: Maybe use sentry instead of this?

apps/deduplication/tasks/indexing.py
```diff
@@ -0,0 +1,148 @@
...
+            threshold=LSHIndex.THRESHOLD,
+            num_perm=LSHIndex.NUM_PERM,
+        )
+        index_obj.index = index
+        index_obj.save()
+    return index_obj
+
+
+@shared_task
+@warn_on_exception(logger)
```
thenav56: This way we might miss errors. Shouldn't we handle them instead?
Also, the sentry will skip `logger.warning`.

thenav56 (changes requested):
Looks good. Few comments

apps/deduplication/models.py
bewakes: Why though? Because integer would take up less storage?

apps/deduplication/models.py
bewakes: Nope, no more. It was used while in deepl-server. Thanks for pointing it out.

apps/deduplication/models.py
bewakes: The index needs to persist that's why it is pickled and stored in db. Another alternative would be to store it in other storages like s3 but I think using db is lot cleaner.

One concern is what if the index gets larger. Postgres column can handle columns upto 2 GB. The index indeed increases linearly in size with ~700 Bytes per document. But the number of leads in a project in DEEP is maximum ~8300 till date, which will make the index size around 7MB which is not that much even if we have around 20K leads per project.

We can also use redis as storage, but we will eventually need to persist data somewhere and having both will add complexity.

apps/deduplication/models.py
bewakes: No, not at all. :D

apps/deduplication/models.py
bewakes: Done.

apps/deduplication/models.py
bewakes: done

apps/deduplication/models.py
bewakes: done.

apps/lead/views.py
bewakes: Done.

apps/deduplication/tasks/indexing.py
bewakes: I've left it as it is while adding `logger.error` instead of `logger.warning`.

apps/deduplication/tasks/indexing.py
bewakes: Changed so that `logger.error` is called.

apps/deduplication/models.py
```diff
@@ -0,0 +1,95 @@
...
+        if (
+            hasattr(self, "index_pickle") and
+            self.index_pickle is not None and
+            self.pickle_version is not None
+        ):
+            supported_formats = pickle.compatible_formats
+            if self.pickle_version not in supported_formats:
+                logger.warn(
+                    "Pickle versions not compatible, setting index to None"
+                )  # noqa
```
bimal125: Which flake8 error are you trying to ignore? we can specify the error code here.

apps/deduplication/models.py
```diff
@@ -0,0 +1,95 @@
...
+        self._index = value
+        self.index_pickle = pickle.dumps(value)
+
+    def __init__(self, *args, **kwargs):
+        self._index_loaded = False
+        super().__init__(*args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        # Set self._index_loaded = False so that next time it is
+        # accessed, it is reloaded
```
bimal125: We can use multiline comment here

apps/lead/serializers.py
```diff
@@ -514,6 +515,13 @@ def create(self, validated_data):
                 connector_lead=lead.connector_lead,
                 source__unified_connector__project=lead.project,
             ).update(already_added=True)
+
+        # Call bg task for indexing. This call is okay to make even if lead
+        # has preview and not ready. This is because, if there is no
+        # preview or lead text, the function will return. And later
+        # when preview is ready after extraction, it will be called
+        # again
```
bimal125: We can use multiline comment here

apps/deduplication/models.py
thenav56: Sounds good. Let's add this as a note as well.

apps/deduplication/models.py
thenav56: Super. Looks like Postgres supports smallint. We can use that https://docs.djangoproject.com/en/3.2/ref/models/fields/#positivesmallintegerfield

utils/decorators.py
```diff
@@ -0,0 +1,10 @@
+def log_error_on_exception(logger, exceptions=[]):
+    """A decorator that just catches exceptions and warns"""
```
thenav56:
```suggestion
    """A decorator that just catches exceptions and logs errors"""
```

apps/lead/receivers.py
```diff
@@ -4,10 +4,16 @@
...
 @receiver(models.signals.post_delete, sender=Lead)
 def unset_already_added_connector_lead(sender, instance, **kwargs):
     if instance.connector_lead is None:
         return
     ConnectorSourceLead.update_aleady_added_using_lead(instance, added=False)
+
+
+@receiver(models.signals.post_delete, sender=Lead)
+def remove_lead_from_index_after_deletion(sender, instance, **kwargs):
+    remove_lead_from_index.delay(instance.id)
```
thenav56: Let's use transaction on_commit as well.

apps/deduplication/tests/test_tasks.py
```diff
@@ -0,0 +1,112 @@
...
+        assert lead.is_indexed is True
+        assert lead.indexed_at is not None
+
+        assert index_obj.index is not None
+        assert lead.id in index_obj.index
+
+    def test_find_and_set_duplicate_leads(self):
+        project = ProjectFactory.create()
+        lead1 = LeadFactory.create(project=project)
+        lead2 = LeadFactory.create(project=project)
```
thenav56: Tip: Factory supports create_batch

apps/deduplication/tests/test_tasks.py
```diff
@@ -0,0 +1,112 @@
...
+from deduplication.tasks.indexing import (
+    process_and_index_lead,
+    get_index_object_for_project,
+    index_lead_and_calculate_duplicates,
+    remove_lead_from_index,
+)
+
+
+@pytest.mark.django_db
+class TestTasks(TestCase):
```
thenav56, line 18: Is there a reason behind not using GraphQLTestCase as base?

NOTE: GraphQLTestCase is the new base case where factories are used instead of auto-fixtures.

apps/deduplication/tasks/indexing.py
```diff
@@ -0,0 +1,152 @@
...
+            threshold=LSHIndex.THRESHOLD,
+            num_perm=LSHIndex.NUM_PERM,
+        )
+        index_obj.index = index
+        index_obj.save()
+    return index_obj
+
+
+@shared_task
+@log_error_on_exception(logger)
```
thenav56: One question here. As we are handling all errors, is this required? Not seeing the use case for this.

apps/deduplication/tasks/indexing.py
```diff
@@ -0,0 +1,152 @@
...
+from project.models import Project
+from deduplication.models import LSHIndex
+from deduplication.utils import get_minhash, insert_to_index
+
+logger = get_task_logger(__name__)
+
+
+def find_and_set_duplicate_leads(index: MinHashLSH, lead: Lead, minhash: LeanMinHash):
+    duplicate_lead_ids = index.query(minhash)
+    duplicate_leads = Lead.objects.filter(pk__in=duplicate_lead_ids)
```
thenav56: Just for habit, let's append `_qs` if the variable is a queryset.

apps/lead/serializers.py
```diff
@@ -514,6 +515,13 @@ def create(self, validated_data):
...
+        index_lead_and_calculate_duplicates.delay(lead.id)
```
thenav56: Let's use transaction on_commit

apps/lead/serializers.py
```diff
@@ -687,13 +695,16 @@ def create(self, data):
...
+            lead = LeadExtraction.save_lead_data(
                 lead,
                 data['text_path'],
-                data.get('images_path', [])[:10],   # TODO: Support for more images, to much image will error.
+                data.get('images_path', [])[:10],   # TODO: Support for more images, too much image will error.
                 data.get('total_words_count'),
                 data.get('total_pages'),
             )
+            # Add to deduplication index
+            index_lead_and_calculate_duplicates(lead.id)
```
thenav56: We will need to delay this? If the response time is> 30 it might fail

apps/lead/tests/test_schemas.py
```diff
@@ -352,6 +353,157 @@ def _query_check(**kwargs):
...
+            query MyQuery ($projectId: ID! $duplicatesOf: ID!) {
+              project(id: $projectId) {
+                leads (duplicatesOf: $duplicatesOf) {
+                  results {
+                    id
+                    title
+                    entriesCount {
+                      total
+                      controlled
+                    }
```
thenav56: Are we using this?

apps/lead/tests/test_schemas.py
```diff
@@ -352,6 +353,157 @@ def _query_check(**kwargs):
...
```
thenav56: Are we using this?

thenav56 (changes requested):

apps/deduplication/models.py
bewakes: I have not used multiline comment because at the top of a function it looks more like a docstring.

apps/lead/serializers.py
bewakes: Done. thanks.

apps/lead/serializers.py
bewakes: Done.

apps/lead/serializers.py
bewakes: Yes. That makes more sense. Done.

apps/lead/tests/test_schemas.py
bewakes: NO. fixed.

apps/lead/tests/test_schemas.py
bewakes: No, fixed.

utils/decorators.py
bewakes: Removed this altogether.

apps/lead/receivers.py
bewakes: Done.

apps/deduplication/tests/test_tasks.py
bewakes: The main reason is this does not use any api calls or gql calls. Just unit tests with db access.

apps/deduplication/tasks/indexing.py
bewakes: Removed.

apps/deduplication/tasks/indexing.py
bewakes: Removed.

apps/deduplication/tasks/indexing.py
bewakes: done.

apps/deduplication/models.py
bewakes: It was not even needed. Thanks for pointing out. Removed.

apps/deduplication/models.py
bewakes: Done.

bewakes (commented):
Resolved issues.

thenav56 (approved):
LGTM

I add a comment in the wrong thread about this https://docs.djangoproject.com/en/3.2/ref/models/fields/#positivesmallintegerfield

bimal125 (approved):

PR #1264, opened 2023-01-22
Fix widget id parse issue

thenav56:
Addresses
- #1263

## Changes

- Fix free widgets id parsing logic
- Remove conditionalWidget logics

NOTE: To fix the `pending` column issue. Temp hack
```bash
# Replace $POSTGRES_USER and $POSTGRES_DB if you are using a different Postgres username or database.
docker-compose exec db bash -xc 'psql -U $POSTGRES_USER $POSTGRES_DB -c "ALTER TABLE export_export ADD COLUMN pending BOOLEAN DEFAULT FALSE;"'
# UNDO
docker-compose exec db bash -xc 'psql -U $POSTGRES_USER $POSTGRES_DB -c "ALTER TABLE export_export DROP COLUMN pending;"'
```

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/apps/export/entries/report_exporter.py
+++ b/apps/export/entries/report_exporter.py
@@ -190,7 +190,7 @@ class WidgetExporter:
 class ReportExporter:
     def __init__(
         self,
-        exporting_widgets=None,
+        exporting_widgets=None,  # eg: ["517", "43", "42"]
         is_preview=False,
         show_lead_entry_id=True,
         show_assessment_data=True,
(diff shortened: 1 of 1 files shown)


apps/export/entries/report_exporter.py
```diff
@@ -459,29 +453,24 @@ def _generate_for_entry_widget_data(self, entry, para):
...
-                            raw_export_data.append(rep)
-                else:
-                    export_datum = {
-                        **each.data.get('common', {}),
-                        **each.data.get('report', {}),
-                    }
-                    if export_datum.get('widget_key') and export_datum['widget_key'] in self.exporting_widgets_keys:
-                        raw_export_data.append(export_datum)
+                export_datum = {
+                    **each.data.get('common', {}),
```
bewakes: I think `data.get('common') or {}` will be safer here, just in case that the value is deliberately set to `None`. If it is guaranteed that it cannot happen, this should be just fine.

thenav56: Sure. Done. There are lots of these though. Maybe we should just use [pydash](https://pydash.readthedocs.io/en/latest/) `pydash.get(`.

bewakes (approved):

Issue #1269, opened 2023-02-09
Make the dates in excel export consistent with DD-MM-YYYY format


udaynwa:
**Reported by José Cobos Romero on Slack**
Link to the reported message : https://datafriendlyspace.slack.com/archives/C03JD71584B/p1675855499463689

Attached reference screenshots:

![Image](https://user-images.githubusercontent.com/72339626/217723215-6296c5df-ad85-421f-be6e-e992e02c28ee.png)
![Image](https://user-images.githubusercontent.com/72339626/217723216-ff7148aa-9af2-4736-8fb0-c33377b3012b.png)

PR #1270, opened 2023-02-09
Use global date format: %d-%m-%Y

thenav56:
Addresses
- https://github.com/the-deep/server/issues/1269

## Changes
- Update visual format to '%d-%m-%Y'

## QA
- Entry Excel export
- Entry Report export
- ARY Export

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/utils/common.py
+++ b/utils/common.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
 import hashlib
-from collections import Counter
-from functools import reduce
 import os
 import re
 import time
@@ -10,7 +8,10 @@ import string
 import tempfile
 import requests
 import logging
-from datetime import timedelta, datetime
+import datetime
+from typing import Union, Optional
+from collections import Counter
+from functools import reduce
 
 from django.utils.hashable import make_hashable
 from django.utils.encoding import force_str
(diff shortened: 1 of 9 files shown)


utils/common.py
```diff
@@ -124,20 +125,22 @@ def get_valid_xml_string(string, escape=True):
     return ''
 
 
-def format_date(date):
-    return date and date.strftime('%d-%m-%Y')
+def deep_date_format(date: Union[datetime.date, datetime.datetime], fallback: Union[str, None] = '') -> Union[str, None]:
```
bewakes: You can use `Optional[str]` which is the prefered shorthand for `Union[str, None]`

k9845 (approved):
Just curious, why are we using this date format `'%d-%m-%Y'`?

utils/common.py
thenav56: Done.

utils/common.py
```diff
@@ -124,20 +125,22 @@ def get_valid_xml_string(string, escape=True):
     return ''
 
 
-def format_date(date):
-    return date and date.strftime('%d-%m-%Y')
+def deep_date_format(date: Union[datetime.date, datetime.datetime], fallback: Optional[str] = '') -> Optional[str]:
+    if date:
```
bewakes, line 132: This statement implies that the date can be `None`. However the date argument signature says it cannot be `None`. Need to add `None` inside the Union.

bewakes (changes requested):
Looks good to me. A minor request for change.

utils/common.py
thenav56: Done

bewakes (approved):

bimal125 (approved):

PR #1323, opened 2023-06-12
Create Graphdoc Documentation in server

RobertKoikai:
Addresses xxxxxx (eg: #1 the-deep/server#1) \
Depends on xxxxxx (eg: #1 the-deep/server#1)

## Changes

* Detailed list or prose of changes
* Breaking changes
* Changes to configurations

- [ x] temporary files, auto-generated files or secret keys
- [ x] n+1 queries
- [ x] flake8 issues
- [ x] `print`
- [ x] typos
- [ x] unwanted comments

## This PR contains valid:

- [x ] tests
- [ x] permission checks (tests here too)
- [ x] translations


deep/urls.py
```diff
@@ -400,6 +402,10 @@ def get_api_path(path):
     admin.site.__class__ = OTPAdminSite
 
 urlpatterns = [
+    #docs paths
+    path('graphql', GraphQLView.as_view(), name='graphql_endpoint'),
```
thenav56: This is not required. We have a custom view for graphql called `CustomGraphQLView`

deep/views.py
```diff
@@ -42,6 +45,11 @@ def get_frontend_url(path=''):
         path=path,
     )
 
+def graphql_docs(request):
+    # For graphene>=3 use schema.graphql_schema
+    # html = graphdoc.to_doc(GraphQLView().schema.graphql_schema)
```
thenav56: Let's remove the comment if not required.

thenav56 (changes requested):
Hi Robert,

The test is breaking, we also need to add the graphdoc package as well.

.env
thenav56, line 1: This is not required in the server repo. We have this in deeper.

deep/views.py
```diff
@@ -42,6 +45,9 @@ def get_frontend_url(path=''):
         path=path,
     )
 
+def graphql_docs(request):
+    html = graphdoc.to_doc(GraphQLView().schema)
```
thenav56: To fix the __proxy__ type issue.
```suggestion
    html = graphdoc.to_doc(str(CustomGraphQLView().schema))
```

thenav56 (changes requested):

deep/urls.py
```diff
@@ -400,6 +402,10 @@ def get_api_path(path):
...
+    path('docs', graphql_docs, name='graphql_docs'),
```
thenav56: Also, let's use `graphql-docs` as path instead of docs

thenav56 (changes requested):

thenav56:
Closed in favour of https://github.com/the-deep/server/pull/1325

PR #1324, opened 2023-06-13
Feature/summary score cna

Rup-Narayan-Rajbanshi:
## Changes
- Add Mutation and query for:
      - Score
      - Cna
      - Summary
- Add Test Case

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations


apps/assessment_registry/admin.py
thenav56, line 1: Let's try to follow this for any admin pages
- Use Autocomplete for any Foreign fields in
  - Form
  - Filters
- Use prefetch if Foreign fields are used in list_display

deep/schema.py
```diff
@@ -22,6 +22,8 @@
 from assisted_tagging import schema as assisted_tagging_schema
 from unified_connector import schema as unified_connector_schema
 from export import schema as export_schema, mutation as export_mutation
+from assessment_registry import mutation as issue_mutation
+from assessment_registry import schema as issue_schema
```
thenav56:
```suggestion
from assessment_registry import mutation as assessment_registry_mutation
from assessment_registry import schema as assessment_registry_schema
```

apps/assessment_registry/models.py
```diff
@@ -322,144 +339,485 @@ class RatingType(models.IntegerChoices):
...
+        sub_sector = self.sub_sector
+
+        if hasattr(self, 'sector'):
+            if hasattr(self, 'sub_sector'):
+                if sub_sector not in Question.QUESTION_SECTOR_SUB_SECTOR_MAP[sector]:
+                    raise ValidationError('Invalid sebsector selected for given sector provided')
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        super().save(*args, **kwargs)
```
thenav56: Why are we doing this here instead of serializers?

apps/assessment_registry/mutation.py
```diff
@@ -6,21 +6,36 @@
...
     'AssessmentRegistryCreateInputType',
     serializer_class=AssessmentRegistrySerializer
 )
+IssueCreateInputType = generate_input_type_for_serializer(
+    'IssueCreateInputType',
+    serializer_class=IssueSerializer
+)
+
+
+class CreateIssue(PsGrapheneMutation):
```
thenav56: Let's use GrapheneMutation for normal mutations

apps/assessment_registry/mutation.py
```diff
@@ -6,21 +6,36 @@
...
+    result = graphene.Field(IssueType)
+    serializer_class = IssueSerializer
+    model = SummaryIssue
+    permissions = []
 
 
 class CreateAssessmentRegistry(PsGrapheneMutation):
     class Arguments:
-        data = AssessmentRegistryCreateInputType(required=True)
+        data = AssessmentRegistryCreateInputType()
```
thenav56: Why remove required=True

apps/assessment_registry/mutation.py
```diff
@@ -30,7 +45,7 @@ class Arguments:
 
 class UpdateAssessmentRegistry(PsGrapheneMutation):
     class Arguments:
-        data = AssessmentRegistryCreateInputType(required=True)
+        data = AssessmentRegistryCreateInputType()
```
thenav56: required=True

apps/assessment_registry/models.py
```diff
@@ -322,144 +339,485 @@ class RatingType(models.IntegerChoices):
...
+    total_people_severly_in_need = models.IntegerField(null=True, blank=True)
+    total_people_critically_in_need = models.IntegerField(null=True, blank=True)
+
+    class Meta:
+        verbose_name = _("SummaryDimmension")
+
+
+class SummaryIssue(models.Model):
+    class SubPillar(models.IntegerChoices):
+        POLITICS = 0, 'Politics'
```
thenav56: Let's not use 0 for any enum values.
Makes it easy to compare values.
eg: if sub_pillar: always work if there is no zero

apps/assessment_registry/serializers.py
```diff
@@ -1,42 +1,134 @@
...
+
+            if data.get('sub_dimmension') is not None:
+                if data.get('sub_dimmension') != data.get('parent').sub_dimmension:
+                    raise serializers.ValidationError("sub_dimmension does not match between child and parent.")
+        if data.get('parent'):
+            hierarchy_level = get_hierarchy_level(data.get('parent'))
+            if hierarchy_level > 2:
+                raise serializers.ValidationError("Cannot create issue more than two level of hierarchy")
+
+        return data
```
thenav56: After using non-zero values for enums

```suggestion
    def validate(self, data):

        sub_pillar = data.get('sub_pillar')
        sub_dimmension = data.get('sub_dimmension')
        parent = data.get('parent')

        if all([sub_pillar, sub_dimmension])
            raise serializers.ValidationError("Cannot select both sub_pillar and sub_dimmension field.")

        if not any([sub_pillar, sub_dimmension])
            raise serializers.ValidationError("Either sub_pillar or sub_dimmension must be selected")

        if parent:
            if sub_pillar and sub_pillar != parent.sub_pillar:
                raise serializers.ValidationError("sub_pillar does not match between parent and child.")

            if sub_dimmension and sub_dimmension != parent.sub_dimmension:
                raise serializers.ValidationError("sub_dimmension does not match between child and parent.")

            hierarchy_level = get_hierarchy_level(parent)
            if hierarchy_level > 2:
                raise serializers.ValidationError("Cannot create issue more than two level of hierarchy")

        return data
```

apps/assessment_registry/schema.py
```diff
@@ -344,23 +359,57 @@ def resolve_score_ratings(root, info, **kwargs):
...
+
+
+class Query():
+    issue = DjangoObjectField(IssueType)
+    issues = DjangoPaginatedListObjectField(
+        IssueListType,
+        pagination=PageGraphqlPagination(
+            page_size_query_param='pageSize'
+        )
+    )
```
thenav56: The term issue is too generic to be used in the root. Let's add a prefix. Something like this
```suggestion
    assessment_reg_summary_issue = DjangoObjectField(AssessmentRegistrySummaryissueType)
    assessment_reg_summary_issues = DjangoPaginatedListObjectField(
        AssessmentRegistrySummaryissueListType,
        pagination=PageGraphqlPagination(
            page_size_query_param='pageSize'
        )
    )
```

apps/assessment_registry/schema.py
```diff
@@ -276,6 +211,80 @@ def resolve_external_link(root, info, **kwargs):
...
+    question = graphene.Field(QuestionType, required=True)
+
+    class Meta:
+        model = Answer
+        fields = ("id", "question", "answer",)
+
+
+class IssueType(DjangoObjectType, UserResourceMixin):
+    sub_pillar = graphene.Field(AssessmentRegistrySummarySubPillarTypeEnum, required=False)
+    sub_pillar_display = graphene.String(required=False)
```
thenav56: Let's use EnumDescription here instead of graphene.String. It should work with custom resolve.

utils/common.py
```diff
@@ -597,3 +597,12 @@ def render_string_for_graphql(text):
...
     return text
+
+
+# Get heirarchy level of a django model.
+def get_hierarchy_level(parent_instance):
+    level = 1
+    while parent_instance.parent:
+        level += 1
+        parent_instance = parent_instance.parent
+    return level
```
thenav56: Let's move this out of common, as this might now work for all cases.
You can define utils.py in assessment_registry instead of just in the same function where this is used.

thenav56 (changes requested):

apps/assessment_registry/models.py
Rup-Narayan-Rajbanshi: We are doing this because Question are added from the admin form not from the endpoints.

thenav56: Sounds good. Let's add a note as well.

apps/assessment_registry/mutation.py
thenav56: Also, let's append a prefix to this mutation as well. CreateIssue is generic.

apps/assessment_registry/models.py
```diff
@@ -195,7 +203,7 @@ class AffectedGroupType(models.IntegerChoices):
     sectors = ArrayField(models.IntegerField(choices=SectorType.choices), default=list)
     protection_info_mgmts = ArrayField(
         models.IntegerField(choices=ProtectionInfoType.choices),
-        default=list
+        blank=True, null=True
```
thenav56: Why not `default=list, blank=True`

utils/graphene/mutation.py
```diff
@@ -288,6 +288,7 @@ class GrapheneMutation(BaseGrapheneMutation):
 
     @classmethod
     def perform_mutate(cls, root, info, **kwargs):
+        print("*******************************")
```
thenav56:
```suggestion
```

thenav56 (changes requested):

utils/graphene/mutation.py
```diff
@@ -289,6 +289,7 @@ class GrapheneMutation(BaseGrapheneMutation):
     @classmethod
     def perform_mutate(cls, root, info, **kwargs):
         data = kwargs['data']
+        print('-->', data)
```
thenav56:
```suggestion
```

apps/assessment_registry/filters.py
```diff
@@ -45,3 +57,12 @@ def filter_assessment_registry(self, qs, name, value):
             Q(id=value) |
             Q(lead__title__icontains=value)
         ).distinct()
+
+
+class IssueGQFilterSet(django_filters.FilterSet):
```
thenav56:
```suggestion
class AssessmentRegistryIssueGQFilterSet(django_filters.FilterSet):
```

apps/assessment_registry/models.py
```diff
@@ -322,144 +345,494 @@ class RatingType(models.IntegerChoices):
...
+        EVENT_SHOCK = 2, 'Event/Shock'
+        DISPLACEMENT = 3, 'Displacement'
+        INFORMATION_AND_COMMUNICATION = 4, 'Information & Communication'
+        HUMANITARIAN_ACCESS = 5, 'Humanitarian Access'
+
+#    assessment_registry = models.ForeignKey(
+#        AssessmentRegistry,
+#        on_delete=models.CASCADE,
+#        related_name='summary',
+#    )
```
thenav56:
```suggestion
```

apps/assessment_registry/mutation.py
```diff
@@ -3,19 +3,38 @@
...
     'AssessmentRegistryCreateInputType',
     serializer_class=AssessmentRegistrySerializer
 )
+AssessmentRegistrySummaryIssueCreateInputType = generate_input_type_for_serializer(
+    'AssessmentRegistrySummaryIssueCreateInputType',
+    serializer_class=IssueSerializer
+)
+
+
+class CreateIssue(GrapheneMutation):
```
thenav56:
```suggestion
class AssessmentRegistryCreateIssue(GrapheneMutation):
```

apps/assessment_registry/schema.py
```diff
@@ -276,6 +206,83 @@ def resolve_external_link(root, info, **kwargs):
...
+        model = SummaryIssue
+        fields = [
+            'id', 'parent', 'label', 'full_label',
+        ]
+
+    @staticmethod
+    def resolve_sub_pillar_display(root, info, **kwargs):
+        if root.sub_pillar is not None:
+            return root.get_sub_pillar_display()
+        return None
```
thenav56: This is the default behaviour
```suggestion
```

apps/assessment_registry/serializers.py
```diff
@@ -92,31 +201,41 @@ class Meta:
...
         return data
+
+#    def create(self, data):
+#        summary_data = data.pop('summary')
+#        instance = super().create(data)
+#        Summary.objects.create(
+#            assessment_registry=instance,
+#            **summary_data,
+#        )
+#        return instance
```
thenav56:
```suggestion
```

thenav56 (changes requested):
Few clean-up issues.

Also, can you check if the existing was are also all done?

'
thenav56, line 1: Clean-up

apps/assessment_registry/schema.py
```diff
@@ -276,6 +222,81 @@ def resolve_external_link(root, info, **kwargs):
...
+
+    class Meta:
+        model = SummarySubPillarIssue
+        fields = [
+            "id", "text", "order", "summary_issue", "lead_preview_text_ref"
+        ]
+
+    @staticmethod
+    def resolve_issue(root, info, **kwargs):
+        return root.summary_issue
```
thenav56: Dataloader missing

thenav56 (changes requested):
Some of the previous comments are still pending.

thenav56 (approved):

PR #1328, opened 2023-07-03
Feature/1327-User search should show results when search for full name

RobertKoikai:
Addresses : User search should show results when search for full name

## Changes

* Update the user filter function to allow search by full name

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/user/filters.py
+++ b/apps/user/filters.py
@@ -43,11 +43,10 @@ class UserGqlFilterSet(django_filters.FilterSet):
                 ~models.Q(groupmembership__group_id=value)
             )
         return qs
-    
+
     def filter_search(self, qs, name, value):
         if value:
             first_name, last_name = value.split(' ', 1) if ' ' in value else (value, '')
-        
         if first_name and last_name:
             qs = qs.filter(
                 models.Q(first_name__icontains=first_name) &
@@ -60,7 +59,6 @@ class UserGqlFilterSet(django_filters.FilterSet):
                 models.Q(email__icontains=value) |
                 models.Q(username__icontains=value)
             )
-
         return qs
 
     @property
--- a/apps/user/filters.py
+++ b/apps/user/filters.py
@@ -1,7 +1,8 @@
 import django_filters
 from django.db import models
-
+from django.db.models.functions import Concat
 from utils.graphene.filters import IDFilter
+from .models import Q
 
 from .models import User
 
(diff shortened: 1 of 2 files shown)


apps/user/filters.py
```diff
@@ -46,6 +46,13 @@ def filter_exclude_usergroup(self, qs, name, value):
 
     def filter_search(self, qs, name, value):
         if value:
+            first_name, last_name = value.split(' ', 1) if ' ' in value else (value, '')
+        if first_name and last_name:
+            qs = qs.filter(
+                models.Q(first_name__icontains=first_name) &
+                models.Q(last_name__icontains=last_name)
+            )
+        else:
```
thenav56: Another approach can be to annotate "first_name last_name" at the query level and then use the value directly to search. This way if there are multiple words, it will also work.

https://docs.djangoproject.com/en/4.2/topics/db/aggregation/#generating-aggregates-for-each-item-in-a-queryset
https://docs.djangoproject.com/en/3.2/ref/models/database-functions/#concat

Also, let's update the test case with this new search logic.

thenav56 (changes requested):
Good work.

apps/user/filters.py
RobertKoikai: @thenav56 ,I have updated the function.
Can you please point me to where the test cases are ?

thenav56: Super. Let's also keep existing filters as well.  cc: @AdityaKhatri 
```
                models.Q(first_name__icontains=value) |
                models.Q(last_name__icontains=value) |
 ```

## Add/Update test
We normally add test cases in the individual app directory. For users it will be inside `apps/user/tests/`, since the changes are related to a **query**, we add test cases to `test_schemas.py` or (`test_queries.py`)
For the user's query, lucky we already have one which is defined here https://github.com/the-deep/server/blob/develop/apps/user/tests/test_schemas.py#L535. Now we will need to make sure that new cases are also tested here.

File: apps/user/filters.py [hunk 2]
```diff
@@ -1,7 +1,8 @@
 import django_filters
 from django.db import models
-
+from django.db.models.functions import Concat
 from utils.graphene.filters import IDFilter
+from .models import Q
```
thenav56: Let's not do this, directly use Q from `django.db.models` which is already imported.

thenav56 (changes requested):

apps/user/tests/test_schemas.py
```diff
@@ -587,7 +587,8 @@ def _query_check(filters, **kwargs):
             ('exclude-project-2', dict(membersExcludeProject=project2.pk), 2, [user1, user3]),
             ('exclude-af-1', dict(membersExcludeFramework=af1.pk), 1, [user3]),
             ('exclude-af-2', dict(membersExcludeFramework=af2.pk), 3, [user, user1, user3]),
-            ('search', dict(search='Guy'), 1, [user]),
+            ('search-1', dict(search='Normal'), 1, [user]),
+            ('search-2', dict(search='Guy'), 1, [user]),
```
thenav56: I think we are still missing test cases related to this issue. Let's add more cases like
Search text:
- Firstname Lastname
- FIrstname
- Lastname
- <spaces>Firstname
- Firstname<spaces>
- Firstname Lastname<spaces>
- Email
- Email(Partial text only)

thenav56 (changes requested):

apps/user/tests/test_schemas.py
RobertKoikai: @thenav56 ,
Is there a difference between the 2nd,4th & 5th Firstname ?
and also 1st & 6th Firstname Lastname ?

thenav56: Hi @RobertKoikai, yes there is but I didn't check the markdown preview. Here is the raw text
```
Search text:
- Firstname Lastname
- FIrstname
- Lastname
- <spaces>Firstname
- Firstname<spaces>
- Firstname Lastname<spaces>
- Email
- Email(Partial text only)
```

thenav56 (approved):

PR #1340, opened 2023-07-26
Make extenal link type from text to URL.

Rup-Narayan-Rajbanshi:
## Changes
- Make extenal link type from text to URL.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [x] translations


apps/assessment_registry/schema.py
```diff
@@ -262,13 +263,20 @@ class Meta:
...
-        fields = ("id", "file", "external_link")
+        fields = ("id", "file")
 
     document_type = graphene.Field(AssessmentRegistryDocumentTypeEnum, required=True)
     document_type_display = EnumDescription(source='get_document_type_display', required=True)
 
+    def resolve_external_link(root, info, **kwargs):
+        if root.external_link == "":
+            return None
+        return root.external_link
```
thenav56: Let's define a utils function. Something like this

utils/common.py
```python3
def render_string_for_graphql(text):
    """
    Return null if text is empty ("")
    """
    if text == '':
        return None
    return text
```

```suggestion
        if render_string_for_graphql(root.external_link)
```

apps/geo/schema.py
```diff
@@ -107,6 +107,13 @@ class Meta:
     region_title = graphene.String(required=True)
     admin_level_title = graphene.String(required=True)
 
+    @staticmethod
+    def resolve_region_title(root, info, **kwargs):
+        return root.admin_level.region.title
+
+    def resolve_admin_level_title(root, info, **kwargs):
+        return root.admin_level.title
+
```
thenav56: This will create an N+1 query.

You can pass the queryset through get_geo_area_queryset_for_project_geo_area_type to annotate the region and admin_level title.

apps/geo/dataloaders.py
```diff
@@ -19,7 +20,21 @@ def batch_load_fn(self, keys):
         return Promise.resolve([_map.get(key) for key in keys])
 
 
+class GeoAreaLoader(DataLoaderWithContext):
+    def batch_load_fn(self, keys):
+        ary_geo_area_qs = AssessmentRegistry.locations.through.objects\
+            .filter(assessmentregistry__in=keys).prefetch_related('geoarea')
```
thenav56: Let's use defer to remove heavy columns like polygons and centroid which are not used.

apps/assessment_registry/models.py
```diff
@@ -288,7 +289,8 @@ class DocumentType(models.IntegerChoices):
         related_name='assessment_reg_file',
         null=True, blank=True
     )
-    external_link = models.TextField(blank=True)
+    external_link = models.URLField(max_length=500, blank=True)
+    file_path = models.CharField(max_length=500, blank=True)
```
thenav56: Why do we need `file_path`?

apps/gallery/serializers.py
```diff
@@ -21,6 +21,12 @@
 FILE_READONLY_FIELDS = ('metadata', 'mime_type',)
 
 
+class FileUploadSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = File
+        fields = ('__all__')
```
thenav56: Can you make sure existing logic are also here like FileSerializer? Maybe we can just create a new class using that instead?

schema.graphql
```diff
@@ -2039,6 +2040,18 @@ type FileFieldType {
   url: String
 }
 
+input FileUploadInputType {
+  clientId: String
+  title: String!
+  file: Upload
```
thenav56: This should be required right?

thenav56 (changes requested):
Some comments

apps/assessment_registry/schema.py
```diff
@@ -262,13 +264,18 @@ class Meta:
 
 
 class AdditionalDocumentType(DjangoObjectType, UserResourceMixin, ClientIdMixin):
+    external_link = graphene.String(required=False)
+
```
thenav56: Let's move this to after Class Meta like others

thenav56 (changes requested):

thenav56 (dismissed):

thenav56 (approved):

thenav56:
Flake8 issues are already fixed in another PR which will be merged to develop soon.

Issue #1349, opened 2023-08-09
Support search for items with accents throughout the platform


subinasr:
### Overview
In SearchSelectInputs, accents like é, ë, ç, ø, etc in search are not supported.

### Problem Statement
 For example, if I search for Organización Nacional Indígena de Colombia (ONIC) and do not include the accent in any of the words, it will not appear in suggestions, so a new one will be created without an accent.

[Trello card](https://trello.com/c/VNCtsp3S/300-the-organization-search-bar-when-adding-documents-to-deep-from-the-extension-or-from-the-app-itself-excludes-words-if-they-do-no)

### Acceptance Criteria
Options containing ligatures (accents) should appear with its base alphabet search.
For eg: the search should consider "Tèst" and "test" as same.

### Screenshots

![Image](https://github.com/the-deep/server/assets/51225380/52840634-38e5-417e-aec5-092c0438377b)
![Image](https://github.com/the-deep/server/assets/51225380/c53e2170-294c-474e-a95a-9096eb2973f6)

PR #1352, opened 2023-08-15
Support search for items with accents in geolocation module

RobertKoikai:
Addresses
- https://github.com/the-deep/server/issues/1349
SearchSelectInputs, accents like é, ë, ç, ø, etc in search are not supported in the geo areas

## Changes

* Update geo_area_label method to allow for search accents e.g é, ë, ç, ø
* Enable unaccent extension in postgreSQL

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [ ] tests
-[ ] translations


apps/user/migrations/0004_auto_20171129_0851.py
```diff
@@ -15,4 +15,5 @@ class Migration(migrations.Migration):
     operations = [
         CreateExtension('pg_trgm'),
         CreateExtension('postgis'),
+        CreateExtension('unaccent'),
```
thenav56: This will not work as this migration is already run in all of the existing systems. You need to create a new one.
I recommend looking at the Django documentation https://docs.djangoproject.com/en/4.2/topics/migrations/ for full context.

apps/geo/filter_set.py
```diff
@@ -87,8 +87,13 @@ class Meta:
             },
         }
 
-
 # ------------------------------ Graphql filters -----------------------------------
+
+
+class Unaccent(models.Func):
+    function = 'UNACCENT'
+
```
thenav56: Not required. Django provides this by default.
https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/lookups/#unaccent

apps/geo/filter_set.py
```diff
@@ -109,7 +114,10 @@ class Meta:
 
     def geo_area_label(self, queryset, _, value):
         if value:
-            return queryset.filter(title__icontains=value)
+            normalized_value = value.lower()
```
thenav56: Not required, `i` in `icontains` does case-insensitive search.
https://docs.djangoproject.com/en/4.2/ref/models/querysets/#icontains

thenav56 (changes requested):
Let's add test cases as well.

apps/geo/filter_set.py
thenav56: @RobertKoikai Can you also try using this instead `unaccent__icontains`?

RobertKoikai:
> 



> 

@thenav56 ,
I think the test cases you provided should still work even with the unaccent functionality is applied.

apps/geo/filter_set.py
RobertKoikai: @thenav56 ,yes this also worked well, plus its short & precise.

thenav56:
> > 
> 
> > 
> 
> @thenav56 , I think the test cases you provided should still work even with the unaccent functionality is applied.

Let's add an accent in the title of the test, this should test this as well.
https://github.com/the-deep/server/blob/develop/apps/geo/tests/test_schemas.py#L61

apps/user/migrations/0004_auto_20171129_0851.py
thenav56: Let's remove this.

RobertKoikai: @thenav56 

> Let's remove this.

This is removed .

thenav56 (approved):

PR #1370, opened 2023-09-22
Add filter is_parent for assessment-registry issues.

Rup-Narayan-Rajbanshi:
## Changes
- Add filter is_parent for assessment registry issues list.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [ ] translations

### DIFF
--- a/apps/assessment_registry/dataloaders.py
+++ b/apps/assessment_registry/dataloaders.py
@@ -39,7 +39,7 @@ class AssessmentRegistryIssueChildLoader(DataLoaderWithContext):
             'parent'
         ).annotate(
             child_count=Count(
-                'parent'
+                'id'
             )
         ).values(
             'parent',
--- a/apps/assessment_registry/dataloaders.py
+++ b/apps/assessment_registry/dataloaders.py
@@ -1,3 +1,6 @@
+from itertools import count
+from operator import countOf
+from django.db import models
 from collections import defaultdict
 from promise import Promise
 
@@ -21,7 +24,26 @@ class AssessmentRegistryOrganizationsLoader(DataLoaderWithContext):
 
 class AssessmentRegistryIssueLoader(DataLoaderWithContext):
     def batch_load_fn(self, keys):
-        qs = SummaryIssue.objects.filter(id__in=keys)
+        qs = SummaryIssue.objects.filter(
+            parent__in=keys
+        ).order_by().values('parent').annotate(
+            count=models.Count('id')
+        ).values_list(
+            'parent', 'count'
+        )
+
+        print("QS*************************", qs)
+        counts = {
+            ass_reg_id: count
+            for ass_reg_id, count in qs
+        }
+
+        return Promise.resolve([counts.get(key, 0) for key in keys])
+
+
+class AssessmentRegistryIssueChildLoader(DataLoaderWithContext):
+    def batch_load_fn(self, keys):
+        qs = SummaryIssue.objects.filter(parent__id__in=keys)
         _map = {
             issue.pk: issue
             for issue in qs
@@ -38,3 +60,7 @@ class DataLoaders(WithContextMixin):
     @cached_property
     def issues(self):
         return AssessmentRegistryIssueLoader(context=self.context)
+
+    @cached_property
+    def child_issues(self):
+        return AssessmentRegistryIssueChildLoader(context=self.context)
--- a/apps/assessment_registry/dataloaders.py
+++ b/apps/assessment_registry/dataloaders.py
@@ -1,10 +1,9 @@
-from itertools import count
-from operator import countOf
-from django.db import models
+from django.db.models import Count
 from collections import defaultdict
 from promise import Promise
 
 from django.utils.functional import cached_property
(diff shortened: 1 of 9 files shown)


apps/assessment_registry/filters.py
thenav56, line 1: Let's add some test for this change as well.

thenav56 (changes requested):

apps/assessment_registry/filters.py
Rup-Narayan-Rajbanshi: Done sir. @thenav56

apps/assessment_registry/filters.py
```diff
@@ -74,3 +75,10 @@ def filter_assessment_registry_issues(self, qs, name, value):
         return qs.filter(
             label__icontains=value
         )
+
+    def filter_is_parent(self, qs, name, value):
+        if not value:
+            return qs
+        return qs.filter(
+            parent__isnull=True
+        )
```
thenav56: I think we may need to do something like this instead to avoid confusion.

```suggestion
        if value is None:
            return qs
        if value:
            # For parent
            return qs.filter(
                parent__isnull=False
            )
        # For child
        return qs.filter(
            parent__isnull=True
        )
```

apps/assessment_registry/tests/test_schemas.py
```diff
@@ -301,3 +301,38 @@ def _query_check(**kwargs):
...
+        member_user = UserFactory.create()
+        self.force_login(member_user)
+
+        parent_issue1, parent_issue2, parent_issue3 = SummaryIssueFactory.create_batch(3, parent=None)
+        child_issue1 = SummaryIssueFactory.create(parent=parent_issue1)
+        child_issue2 = SummaryIssueFactory.create(parent=parent_issue2)
+
+        for filter_data, expected_issues in [
+            ({'isParent': True}, [parent_issue1, parent_issue2, parent_issue3]),
+            ({'isParent': False}, [child_issue1, child_issue2, parent_issue1, parent_issue2, parent_issue3]),
```
thenav56: Related to above comment
```suggestion
            ({}, [child_issue1, child_issue2, parent_issue1, parent_issue2, parent_issue3]),
            ({'isParent': True}, [cparent_issue1, parent_issue2, parent_issue3]),
            ({'isParent': False}, [child_issue1, child_issue2]),
```

thenav56 (changes requested):

apps/assessment_registry/schema.py
```diff
@@ -270,6 +273,12 @@ class Meta:
             'full_label',
         ]
 
+    def resolve_child_count(root, info, **kwargs):
+        return SummaryIssue.objects.filter(parent=root).count()
```
thenav56: Let's use dataloader

apps/assessment_registry/schema.py
```diff
@@ -270,6 +273,12 @@ class Meta:
...
+
+    def resolve_level(root, info, **kwargs):
+        return get_hierarchy_level(root)
```
thenav56: N+1 issue

```SQL
WITH RECURSIVE parents AS (
        SELECT
            sub_g.id,
            sub_g.id as main_entity_id,
            sub_g.parent_id
        FROM {Model._meta.db_table} sub_g
        WHERE
            sub_g.id in %ids
    UNION
        SELECT
            sub_parent_g.id,
            parents.main_entity_id,
            sub_parent_g.parent_id
        FROM {Model._meta.db_table} AS sub_parent_g
            INNER JOIN parents ON sub_parent_g.id = parents.parent_id
) SELECT
    main_entity_id,
    count(*)
FROM parents
GROUP BY main_entity_id
```

thenav56 (changes requested):

apps/assessment_registry/dataloaders.py
```diff
@@ -29,6 +31,63 @@ def batch_load_fn(self, keys):
...
+class AssessmentRegistryIssueChildLoader(DataLoaderWithContext):
+    def batch_load_fn(self, keys):
+        qs = SummaryIssue.objects.filter(
+            parent__in=keys
+        ).values(
+            'parent'
+        ).annotate(
+            child_count=Count(
+                'parent'
+            )
```
thenav56: Can you confirm if this is working?
We will need to count 'id' instead of 'parent', I think right now all value are either 0 or 1.
```suggestion
            child_count=Count(
                'id'
            )
```

thenav56 (changes requested):
Let's update the GraphQL schema as well.

apps/assessment_registry/dataloaders.py
Rup-Narayan-Rajbanshi: @thenav56 dai,  Both code is working fine.

thenav56: By working, can you check if the values are correct?

thenav56 (dismissed):

Rup-Narayan-Rajbanshi:
@thenav56 thank you.

PR #1417, opened 2024-01-31
Update created_at in ary data migration to old created_at

Rup-Narayan-Rajbanshi:
## Changes
- Add old created_at date for new assessment_registry created_at date during data migration.

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations


apps/assessment_registry/models.py
thenav56, line 1: Could you also clean this up?
https://github.com/the-deep/server/blob/96917b82f73f8d2bf212d5b733a09b56c94fcfa6/apps/assessment_registry/models.py#L153-L161

Also, is there any reason why we have nullable lead?

apps/assessment_registry/management/commands/migrate_old_assessments.py
```diff
@@ -332,6 +332,8 @@ def _get_bg_crisis_type():
 
         }
         assessment_reg, created = AssessmentRegistry.objects.get_or_create(**input_data)
+        assessment_reg.created_at = old_ary.created_at
+        assessment_reg.save()
```
thenav56: Let's use `.update` instead of this. We still need to use the same data for updated_at as well.

Maybe we can copy all created/updated information from existing ARY to new ones as a last step in the migration script.
```python
AssessmentRegistry.objects.update(
    created_at=models.F('lead__assessment__created_at'),
    created_by=models.F('lead__assessment__created_by'),
    updated_at=models.F('lead__assessment__updated_at'),
    updated_by=models.F('lead__assessment__updated_by'),
)
```

thenav56 (changes requested):

apps/assessment_registry/models.py
Rup-Narayan-Rajbanshi: @thenav56 I think it was a mistake during solving merge conflct. I have resove it in the latest push.

apps/assessment_registry/management/commands/migrate_old_assessments.py
```diff
@@ -332,6 +332,15 @@ def _get_bg_crisis_type():
...
         assessment_reg, created = AssessmentRegistry.objects.get_or_create(**input_data)
+        assessment_reg.save()
+
+        # Update the new assessment's create date, update date to old date
+        AssessmentRegistry.objects.update(
+            created_at=models.F('lead__assessment__created_at'),
+            created_by=models.F('lead__assessment__created_by'),
+            updated_at=models.F('lead__assessment__updated_at'),
+            updated_by=models.F('lead__assessment__updated_by'),
+        )
```
thenav56: Let's use this like `migrate_cna_questions()`. Running this for every Assessment Registry copy doesn't make sense.

thenav56 (changes requested):

apps/assessment_registry/management/commands/migrate_old_assessments.py
```diff
@@ -332,6 +342,7 @@ def _get_bg_crisis_type():
 
         }
         assessment_reg, created = AssessmentRegistry.objects.get_or_create(**input_data)
+        assessment_reg.save()
```
thenav56: We don't need this right?

thenav56 (changes requested):

thenav56 (approved):

thenav56 (approved):

PR #1431, opened 2024-03-13
Fix/topic model callback

sudan45:
## Changes
* TopicModel callback response has been changed
* Add Analytical Statement title from NLP while AutoClustering 

- [ ] n+1 queries
- [ ] flake8 issues
- [ ] `print`
- [ ] typos
- [ ] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations


apps/analysis/models.py
```diff
@@ -388,6 +388,7 @@ def __str__(self):
 
 class AnalyticalStatement(UserResource):
     statement = models.TextField()
+    title = models.CharField(max_length=150, null=True, blank=True)
```
thenav56: Let's use this... same for TopicModelCluster as well. Maybe let's use `title` there as well.  Just for consistency.
```suggestion
    title = models.CharField(max_length=150, blank=True)
```

thenav56 (changes requested):
Looks good. Minor suggestions.

Also, let's re-create or squash migrations.

thenav56 (approved):

Issue #1439, opened 2024-03-27
Implement graphql  query node for assignment 


sudan45:
### Problem Statement
The 'Assignment' features uses a Rest API.

### Acceptance Criteria
The 'Assignment' feature works with a graphql query.

#### serializers
https://github.com/the-deep/server/blob/bbfdef24fa80f04a046cca7f957297adc534ba15/apps/notification/serializers.py#L72-L93

#### views
https://github.com/the-deep/server/blob/bbfdef24fa80f04a046cca7f957297adc534ba15/apps/notification/views.py#L80-L103

#### testcases
https://github.com/the-deep/server/blob/bbfdef24fa80f04a046cca7f957297adc534ba15/apps/notification/tests/test_apis.py#L235-L266
Rest Framework urls 
``` 
method: GET
url: /api/v1/assignments/
request params : ?is_done=3&offset=0&limit=5
response: [
        {
            "id": ID,
            "createdAt": Datetime,
            "projectDetails": ProjectDetails
            "createdByDetails": UserDetails
            "contentObjectDetails": ContentDetails(ie lead, entry details)
            "contentType": ID,
            "isDone": Boolean,
            "contentObjectType": str
        }
]
```
Expected graphql 

```
enum AssignmentContentTypeEnum {
  LEAD
  ENTRY_REVIEW_COMMENT
}

type AssignmentContentDataType {
  contentType: AssignmentContentTypeEnum
  lead: AssignmentLeadDetailType
  entryReviewComment: AssignmentEntryReviewCommentDetailType
}

input AssignmentInputType {
  isDone: Boolean
}

type AssignmentLeadDetailType {
  id: ID!
  title: String!
}

type AssignmentProjectDetailType {
  id: ID!
  title: String!
}

type AssignmentType {
  id: ID!
  createdAt: DateTime!
  createdBy: UserType
  createdFor: UserType!
  project: AssignmentProjectDetailType
  isDone: Boolean!
  objectId: Int!
  contentData: AssignmentContentDataType
}

type AssignmentListType {
  results: [AssignmentType!]
  totalCount: Int
  page: Int
  pageSize: Int
}

type AssignmentUpdate {
  errors: [GenericScalar!]
  ok: Boolean
  result: AssignmentType
}
```

sudan45:
@puranban @AdityaKhatri @subinasr 
Note: The client needs to review the lead and entry content types. If contenttype is lead, map data with LeadType; if not, map with EntryType.

PR #1440, opened 2024-03-27
implement graphql query and mutation node for assignment

sudan45:
Addresses
- https://github.com/the-deep/server/issues/1439

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/apps/notification/dataloaders.py
+++ b/apps/notification/dataloaders.py
@@ -2,9 +2,10 @@ from django.utils.functional import cached_property
 
 from promise import Promise
 
-from entry.models import Entry
+from entry.models import EntryComment
 from notification.models import Assignment
 from lead.models import Lead
+from quality_assurance.models import EntryReviewComment
 
 from utils.graphene.dataloaders import DataLoaderWithContext, WithContextMixin
 
@@ -14,32 +15,50 @@ class AssignmentLoader(DataLoaderWithContext):
         assignment_qs = list(
             Assignment.objects
             .filter(id__in=keys)
-            .values_list('id', 'content_type__app_label', 'object_id')
+            .values_list('id', 'content_type__model', 'object_id')
         )
 
         leads_id = []
-        entries_id = []
+        entry_review_comment_id = []
+
         for _, content_type, object_id in assignment_qs:
-            if content_type == 'entry':
-                entries_id.append(object_id)
-            elif content_type == 'lead':
+            # TODO use dict map function
+            if content_type == Lead._meta.model_name:
                 leads_id.append(object_id)
+            elif content_type == EntryReviewComment._meta.model_name:
+                entry_review_comment_id.append(object_id)
 
-        lead_id_map = {
-            lead.id: lead
-            for lead in Lead.objects.filter(id__in=leads_id)
-        } if leads_id else {}
+        _lead_id_map = {}
 
-        entry_id_map = {
-            entry.id: entry
-            for entry in Entry.objects.filter(id__in=entries_id)
-        } if entries_id else {}
+        for _id, title in Lead.objects.filter(id__in=leads_id).values_list('id', 'title'):
+            _lead_id_map[_id] = dict(
+                id=_id,
+                title=title
+            )
+
+        _entry_review_comment_id_map = {}
+
+        for _id, text, entry_id, entry_excerpt, lead_id in EntryReviewComment.objects.filter(
+            id__in=entry_review_comment_id).values_list(
+                'id',
+                'comment_texts__text',
+                'entry__id',
+                'entry__excerpt',
+                'entry__lead_id'
+        ):
+            _entry_review_comment_id_map[_id] = dict(
+                id=_id,
+                text=text,
+                entry_id=entry_id,
+                entry_excerpt=entry_excerpt,
+                lead_id=lead_id
+            )
 
         _result = {
             _id: {
                 'content_type': content_type,
-                'lead': lead_id_map.get(object_id) if content_type == 'lead' else None,
-                'entry': entry_id_map.get(object_id) if content_type == 'entry' else None,
+                'lead': _lead_id_map.get(object_id) if content_type == 'lead' else None,
+                'entry_review_comment': _entry_review_comment_id_map.get(object_id) if content_type == 'entryreviewcomment' else None,
             }
             for _id, content_type, object_id in assignment_qs
         }
--- a/apps/notification/dataloaders.py
+++ b/apps/notification/dataloaders.py
@@ -1,4 +1,5 @@
 from django.utils.functional import cached_property
+from django.db import models
 
 from promise import Promise
 
@@ -7,6 +8,10 @@ from lead.models import Lead
 from quality_assurance.models import EntryReviewComment
 
 from utils.graphene.dataloaders import DataLoaderWithContext, WithContextMixin
+
+
+def get_model_name(model: models.Model) -> str:
+    return model._meta.model_name
 
 
 class AssignmentLoader(DataLoaderWithContext):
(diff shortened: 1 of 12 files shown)


apps/notification/mutation.py
```diff
@@ -44,5 +44,37 @@ def mutate(root, info, data):
...
+        return cls(result=instance, ok=True)
+
+
+class AsssignmentBulkStatusUpdate(PsGrapheneMutation):
+    class Arguments:
+        is_done = graphene.Boolean(required=True)
+    model = Assignment
+    result = graphene.List(graphene.NonNull(AssignmentType))
+    permissions = []
+
```
Rup-Narayan-Rajbanshi: Lets check for unauthenticated user.

apps/notification/mutation.py
```diff
@@ -44,5 +44,37 @@ def mutate(root, info, data):
...
+    model = Assignment
+    result = graphene.Field(AssignmentType)
+    permissions = []
+
+    @classmethod
+    def perform_mutate(cls, root, info, **kwargs):
+        instance = cls.model.objects.get(id=kwargs['id'])
+        instance.is_done = kwargs['is_done']
+        instance.save(update_fields=['is_done'])
+        return cls(result=instance, ok=True)
```
thenav56: For update let's do a regular mutation
```suggestion
class AssignmentUpdate(GrapheneMutation):
    class Arguments:
        id = graphene.ID(required=True)
        data = AssignmentInputType(required=True)

    model = Assignment
    serializer_class = AssignmentSerializer
    result = graphene.Field(AssignmentType)
    permissions = []

    @classmethod
    def filter_queryset(cls, qs, info):
        return Assignment.get_for(user)
```

apps/notification/mutation.py
```diff
@@ -44,5 +44,37 @@ def mutate(root, info, data):
...
+        is_done = graphene.Boolean(required=True)
+    model = Assignment
+    result = graphene.List(graphene.NonNull(AssignmentType))
+    permissions = []
+
+    @classmethod
+    def perform_mutate(cls, root, info, **kwargs):
+        instance = cls.model.objects.filter(created_for=info.context.user, is_done=False)
+        instance.update(is_done=kwargs['is_done'])
+        return cls(result=instance, ok=True)
```
thenav56: Looking at the codebase/UI we only need action for is_done=True, Let's do that only.

Also, we don't need to return anything, as this will update a lot of objects.

Also, PsGrapheneMutation is for entities related to project. Use GrapheneMutation here

AsssignmentMarkAllAsReadUpdate should be fine for this.

apps/notification/mutation.py
```diff
@@ -44,5 +44,37 @@ def mutate(root, info, data):
...
+class AsssignmentBulkStatusUpdate(PsGrapheneMutation):
+    class Arguments:
+        is_done = graphene.Boolean(required=True)
+    model = Assignment
+    result = graphene.List(graphene.NonNull(AssignmentType))
+    permissions = []
+
+    @classmethod
+    def perform_mutate(cls, root, info, **kwargs):
+        instance = cls.model.objects.filter(created_for=info.context.user, is_done=False)
```
thenav56: We already have a function for this. Assignment.get_for(user)

apps/notification/schema.py
```diff
@@ -24,6 +29,13 @@ def get_user_notification_qs(info):
     )
 
 
+def get_user_assignment_qs(info):
+    track_user(info.context.request.user.profile)
+    return Assignment.objects.filter(
+        created_for=info.context.request.user
+    )
```
thenav56: We already have a function for this. Assignment.get_for(user)

apps/notification/schema.py
```diff
@@ -55,6 +100,12 @@ class Query:
             page_size_query_param='pageSize'
         )
     )
+    assignment = DjangoPaginatedListObjectField(
```
thenav56:
```suggestion
    assignments = DjangoPaginatedListObjectField(
```

apps/notification/schema.py
```diff
@@ -41,12 +53,45 @@ def get_custom_queryset(queryset, info, **kwargs):
...
+    def resolve_content_type(root, info):
+        return root.content_type.model
+
+    @staticmethod
+    def resolve_lead_type(root, info):
+        return Lead(root.object_id) if root.content_type.model == 'lead' else None
+
+    @staticmethod
+    def resolve_entry_type(root, info):
+        return Entry(root.object_id) if root.content_type.model == 'entry' else None
```
thenav56: All this data are from another table.

We need to define data loaders for these.

The easiest way will be to create a nested object with `content_type,lead_type,entry_type`. And define the data loader at **AssignmentType** so that the data loader can `content_type,lead_type,entry_type` for each assignment with single data loader.

thenav56 (changes requested):

apps/notification/mutation.py
sudan45: global permisson is always true @Rup-Narayan-Rajbanshi

thenav56 (changes requested):
We still have issues here.

We need to handle all of this for content types
```
Lead: AssignmentLeadSerializer()
EntryComment: AssignmentEntryCommentSerializer()
EntryReviewComment: AssignmentEntryReviewCommentSerializer()
```

Also, let's define enums for content types as well for better type support.

apps/notification/dataloaders.py
```diff
@@ -0,0 +1,71 @@
...
+
+        _entry_review_comment_id_map = {}
+
+        for _id, text, entry_id, entry_excerpt, lead_id in EntryReviewComment.objects.filter(
+            id__in=entry_review_comment_id).values_list(
+                'id',
+                'comment_texts__text',
+                'entry__id',
+                'entry__excerpt',
+                'entry__lead_id'
```
thenav56: Let's clean up fields we have removed from schema

apps/notification/dataloaders.py
```diff
@@ -0,0 +1,71 @@
...
+            Assignment.objects
+            .filter(id__in=keys)
+            .values_list('id', 'content_type__model', 'object_id')
+        )
+
+        leads_id = []
+        entry_review_comment_id = []
+
+        for _, content_type, object_id in assignment_qs:
+            # TODO use dict map function
```
thenav56: I forgot about this TODO. Do you remember what we need here?

apps/notification/dataloaders.py
```diff
@@ -0,0 +1,71 @@
...
+                entry_excerpt=entry_excerpt,
+                lead_id=lead_id
+            )
+
+        _result = {
+            _id: {
+                'content_type': content_type,
+                'lead': _lead_id_map.get(object_id) if content_type == Lead._meta.model else None,
+                'entry_review_comment':
+                _entry_review_comment_id_map.get(object_id) if content_type == EntryReviewComment._meta.model else None,
```
thenav56: Easy read :D 
```suggestion
                'content_type': content_type,
                'lead': (
                    _lead_id_map.get(object_id)
                    if content_type == Lead._meta.model else None
                ),
                'entry_review_comment': (
                    _entry_review_comment_id_map.get(object_id)
                    if content_type == EntryReviewComment._meta.model else None
                ),
```

thenav56 (changes requested):
Looks good. Few changes required.

I am looking at the test fail issue.

apps/notification/tests/test_mutation.py
```diff
@@ -48,3 +51,83 @@ def _query_check(minput, **kwargs):
...
+        self.assertEqual(len(assignments_qs), 0)
+
+    def test_individual_assignment_update_status(self):
+        self.indivdual_assignment_query = '''
+            mutation Mutation($isDone: Boolean, $id: ID! ){
+                assignmentUpdate(id: $id, data: {isDone: $isDone}){
+                    ok
+                    errors
+                    result{
+                        id,
```
thenav56:
```suggestion
                        id
```

thenav56 (approved):
Let's fix the last issue. Looks good now

PR #1461, opened 2024-04-04
add query of project roles

sudan45:
Addresses:
- https://github.com/the-deep/server/issues/1485

- [ ] n+1 queries
- [ ] flake8 issues
- [ ] `print`
- [ ] typos
- [ ] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/apps/project/schema.py
+++ b/apps/project/schema.py
@@ -592,6 +592,7 @@ class Query:
     recent_projects = graphene.List(graphene.NonNull(ProjectDetailType))
     recent_activities = graphene.List(graphene.NonNull(RecentActivityType))
     project_explore_stats = graphene.Field(ProjectExploreStatType)
+    project_roles = graphene.List(graphene.NonNull(ProjectRoleType), required=True)
 
     # only the region for which project are public
     projects_by_region = DjangoPaginatedListObjectField(ProjectByRegionListType)
@@ -650,3 +651,7 @@ class Query:
     @staticmethod
     def resolve_user_pinned_project(root, info, **kwargs):
         return ProjectPinned.objects.filter(user=info.context.user)
+
+    @staticmethod
+    def resolve_project_roles(root, info) -> QuerySet:
+        return ProjectRole.objects.all()
--- a/apps/project/tests/test_schemas.py
+++ b/apps/project/tests/test_schemas.py
@@ -15,6 +15,7 @@ from project.models import (
     ProjectUserGroupMembership,
     ProjectStats,
     Project,
+    ProjectRole,
 )
 from deep.permissions import ProjectPermissions as PP
 from deep.caches import CacheKey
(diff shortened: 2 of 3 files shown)


apps/project/schema.py
```diff
@@ -650,3 +655,7 @@ def resolve_public_projects_by_region(*args, **kwargs):
     @staticmethod
     def resolve_user_pinned_project(root, info, **kwargs):
         return ProjectPinned.objects.filter(user=info.context.user)
+
+    @staticmethod
+    def resolve_project_role(root, info):
```
k9845:
```suggestion
    def resolve_project_role(root, info) -> QuerySet:
```

k9845 (commented):
@sudan45 Let's also specify issue in pr.

apps/project/schema.py
```diff
@@ -592,6 +596,7 @@ class Query:
     recent_projects = graphene.List(graphene.NonNull(ProjectDetailType))
     recent_activities = graphene.List(graphene.NonNull(RecentActivityType))
     project_explore_stats = graphene.Field(ProjectExploreStatType)
+    project_role = graphene.List(graphene.NonNull(ProjectRoleType))
```
thenav56:
```suggestion
    project_roles = graphene.List(graphene.NonNull(ProjectRoleType))
```

apps/project/schema.py
```diff
@@ -202,6 +202,10 @@ class Meta:
 
     type = graphene.Field(ProjectRoleTypeEnum, required=True)
 
+    @classmethod
+    def queryset(cls):
+        return ProjectRole.objects.all()
+
```
thenav56: This is not required. We use queryset for permission. For normal field this is not required.

apps/project/schema.py
```diff
@@ -650,3 +655,7 @@ def resolve_public_projects_by_region(*args, **kwargs):
     @staticmethod
     def resolve_user_pinned_project(root, info, **kwargs):
         return ProjectPinned.objects.filter(user=info.context.user)
+
+    @staticmethod
+    def resolve_project_role(root, info):
+        return ProjectRoleType.queryset()
```
thenav56: A basic ProjectRole.objects.all() should work here

thenav56 (changes requested):

apps/project/schema.py
```diff
@@ -592,6 +592,7 @@ class Query:
     recent_projects = graphene.List(graphene.NonNull(ProjectDetailType))
     recent_activities = graphene.List(graphene.NonNull(RecentActivityType))
     project_explore_stats = graphene.Field(ProjectExploreStatType)
+    project_roles = graphene.List(graphene.NonNull(ProjectRoleType))
```
thenav56:
```suggestion
    project_roles = graphene.List(graphene.NonNull(ProjectRoleType), required=True)
```

apps/project/tests/test_schemas.py
```diff
@@ -1003,6 +1004,29 @@ def _query_check(project_id, **kwargs):
...
+        user = UserFactory.create()
+        # without login
+
+        self.query_check(query, assert_for_error=True)
+
+        # with login
+        self.force_login(user)
+        projectrole = ProjectRole.objects.count()
+        content = self.query_check(query)
+        self.assertEqual(len(content['data']['projectRoles']), projectrole)
```
thenav56:
```suggestion
        project_role_count = ProjectRole.objects.count()
        
        # without login
        self.query_check(query, assert_for_error=True)

        # with login
        self.force_login(user)
        content = self.query_check(query)
        self.assertEqual(len(content['data']['projectRoles']), project_role_count)
```

thenav56 (changes requested):
Look good.

apps/project/tests/test_schemas.py
susilnem: @thenav56 Can we make a factory for Roles and use that here?

thenav56: We have a static set of roles. For test, we have this function which generates here https://github.com/the-deep/server/blob/6a823454917cb15b7df7f4ee33552eaafe8bf027/utils/graphene/tests.py#L137-L189

thenav56 (dismissed):

PR #1480, opened 2024-04-12
Fix/assisted tagging geo loader

sudan45:
## Changes

* add dataloader in draftentry geoarea
* add dataloader in draftentry predication tags

- [ ] n+1 queries
- [ ] flake8 issues
- [ ] `print`
- [ ] typos
- [ ] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations

### DIFF
--- a/apps/assisted_tagging/dataloaders.py
+++ b/apps/assisted_tagging/dataloaders.py
@@ -1,0 +1,24 @@
+from collections import defaultdict
+from promise import Promise
+
+from django.utils.functional import cached_property
+
+from assisted_tagging.models import AssistedTaggingPrediction
+
+from utils.graphene.dataloaders import DataLoaderWithContext, WithContextMixin
+
+
+class DraftEntryPredicationsLoader(DataLoaderWithContext):
+    def batch_load_fn(self, keys):
+        assisted_tagging_qs = AssistedTaggingPrediction.objects\
+            .filter(draft_entry_id__in=keys, is_selected=True)
+        _map = defaultdict(list)
+        for assisted_tagging in assisted_tagging_qs:
+            _map[assisted_tagging.draft_entry_id].append(assisted_tagging)
+        return Promise.resolve([_map.get(key, []) for key in keys])
+
+
+class DataLoaders(WithContextMixin):
+    @cached_property
+    def draft_entry_predications(self):
+        return DraftEntryPredicationsLoader(context=self.context)
--- a/apps/assisted_tagging/schema.py
+++ b/apps/assisted_tagging/schema.py
@@ -10,7 +10,6 @@ from deep.permissions import ProjectPermissions as PP
 
 from geo.schema import (
     ProjectGeoAreaType,
-    get_geo_area_queryset_for_project_geo_area_type,
 )
 from utils.graphene.fields import DjangoPaginatedListObjectField
 from utils.graphene.pagination import NoOrderingPageGraphqlPagination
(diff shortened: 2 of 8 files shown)


apps/assisted_tagging/dataloaders.py
```diff
@@ -0,0 +1,23 @@
...
+from django.utils.functional import cached_property
+from assisted_tagging.models import AssistedTaggingPrediction
+
+from utils.graphene.dataloaders import DataLoaderWithContext, WithContextMixin
+
+
+class DraftEntryPredicationsLoader(DataLoaderWithContext):
+    def batch_load_fn(self, keys):
+        assisted_tagging_qs = AssistedTaggingPrediction.objects\
+            .filter(draft_entry_id__in=keys, is_selected=True).only('draft_entry_id')
```
thenav56: We don't need only here, as we need the whole prediction object

thenav56: We don't need only here, as we need the whole prediction object

apps/assisted_tagging/dataloaders.py
```diff
@@ -0,0 +1,23 @@
...
+        _map = defaultdict(list)
+        for assisted_tagging in assisted_tagging_qs:
+            _map[assisted_tagging.draft_entry.id].append(assisted_tagging)
```
thenav56: Let's not introduce N+1 when trying to fix it
```suggestion
            _map[assisted_tagging.draft_entry_id].append(assisted_tagging)
```

apps/assisted_tagging/schema.py
```diff
@@ -104,19 +103,7 @@ def get_draft_entry_qs(info):  # TODO use dataloader
...
         return qs.prefetch_related(
-            Prefetch(
-                'predictions',
-                queryset=AssistedTaggingPrediction.objects.filter(is_selected=True).order_by('id'),
-            ),
-            Prefetch(
-                'related_geoareas',
-                queryset=get_geo_area_queryset_for_project_geo_area_type().order_by('id'),
-            ),
             'predictions__model_version',
```
thenav56: Let's move this to new dataloader logic

thenav56 (changes requested):

apps/assisted_tagging/schema.py
sudan45: we have already  added logic  in geo/loaders

thenav56 (approved):

susilnem (approved):

apps/assisted_tagging/schema.py
```diff
@@ -159,14 +140,9 @@ class Meta:
...
-    @staticmethod
-    def resolve_wrong_prediction_reviews(root, info, **kwargs):
-        return root.wrong_prediction_reviews.all()   # NOTE: Prefetched by DraftEntry
-
-    @staticmethod
-    def resolve_model_version_deepl_model_id(root, info, **kwargs):
-        return root.model_version.model.model_id   # NOTE: Prefetched by DraftEntry
+    '''
+    NOTE: model_version_deepl_model_id and wrong_prediction_review is not include here because they are not used in frontend
+    '''
```
k9845:
```suggestion
    '''
    NOTE: model_version_deepl_model_id and wrong_prediction_review are not included here because they are not used in frontend
    '''
```

apps/assisted_tagging/dataloaders.py
```diff
@@ -0,0 +1,23 @@
+from collections import defaultdict
+
+from promise import Promise
+from django.utils.functional import cached_property
+from assisted_tagging.models import AssistedTaggingPrediction
+
+from utils.graphene.dataloaders import DataLoaderWithContext, WithContextMixin
```
k9845:
```suggestion
from collections import defaultdict
from promise import Promise

from django.utils.functional import cached_property

from assisted_tagging.models import AssistedTaggingPrediction
from utils.graphene.dataloaders import DataLoaderWithContext, WithContextMixin
```

k9845 (changes requested):

k9845 (approved):

PR #1490, opened 2024-05-16
Fix pinned projects appearing for other user

sudan45:
Addresses 
 - https://github.com/the-deep/server/pull/1489

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [ ] tests
- [ ] translations


susilnem (approved):

apps/project/tests/test_schemas.py
```diff
@@ -1031,6 +1031,42 @@ def test_project_role(self):
...
+            userPinnedProjects {
+                clientId
+                id
+                order
+            }
+        }
+        '''
+
+        user = UserFactory.create()
+        non_member_user = UserFactory.create()
```
thenav56: We aren't checking any membership logic here, let's use proper naming.

```suggestion
        user1 = UserFactory.create()
        user2 = UserFactory.create()
```

apps/project/tests/test_schemas.py
```diff
@@ -1031,6 +1031,42 @@ def test_project_role(self):
         content = self.query_check(query)
         self.assertEqual(len(content['data']['projectRoles']), project_role_count)
 
+    def test_user_pinned_projects_query(self):
+        query = '''
+            query MyQuery {
```
thenav56, line 1036: Let's try to use all the fields we can use for pin projects.

thenav56 (changes requested):

Issue #1501, opened 2024-05-21
Restrict Auto-clustering and Summarization for private projects and confidential leads


subinasr:
Make sure to only allow users to trigger these following mutations if the project is public.
Make sure to filter entries from confidential leads before sending it to NLP.

![Image](https://github.com/the-deep/server/assets/7893887/3e878624-b22b-44ea-98d4-6fa4e6409819)
![Image](https://github.com/the-deep/server/assets/7893887/1ea9cc8c-b7a2-4cfe-a83c-69289ec80168)

PR #1504, opened 2024-06-06
Validation for private projects and confidential leads 

susilnem:
Addresses:
- https://github.com/the-deep/server/issues/1501

## Changes
- Add validation for private projects on Topic model and Automatic summary
- Add filter lead confidential before sending on NLPServer on both topic model and automatic summary

- [x] n+1 queries
- [x] flake8 issues
- [x] `print`
- [x] typos
- [x] unwanted comments

## This PR contains valid:

- [x] tests
- [x] translations

### DIFF
--- a/apps/analysis/serializers.py
+++ b/apps/analysis/serializers.py
@@ -26,7 +26,6 @@ from user.serializers import NanoUserSerializer
 from entry.serializers import SimpleEntrySerializer
 from entry.filter_set import EntryGQFilterSet, EntriesFilterDataInputType
 
-from lead.models import Lead
 from .models import (
     Analysis,
     AnalysisPillar,
@@ -506,8 +505,6 @@ class AnalysisTopicModelSerializer(UserResourceSerializer, serializers.ModelSeri
         return analysis_pillar
 
     def validate_additional_filters(self, additional_filters):
-        # NOTE: overlapping the lead_confidentialities filter, only public leads are allowed
-        additional_filters['lead_confidentialities'] = [Lead.Confidentiality.UNPROTECTED]
         filter_set = EntryGQFilterSet(data=additional_filters, request=self.context['request'])
         if not filter_set.is_valid():
             raise serializers.ValidationError(filter_set.errors)
@@ -564,11 +561,11 @@ class AnalysisAutomaticSummarySerializer(EntriesCollectionNlpTriggerBaseSerializ
     trigger_task_func = trigger_automatic_summary
     widget_tags = StringListField()
 
-    def validate_project(self, project):
+    def validate(self, data):
         project = self.context['request'].active_project
         if project.is_private:
             raise serializers.ValidationError('Automatic summary is not allowed for private projects')
-        return project
+        return data
 
     class Meta:
         model = AutomaticSummary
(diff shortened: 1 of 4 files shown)


apps/analysis/tasks.py
```diff
@@ -51,6 +52,7 @@ def trigger_automatic_summary(_id):
         Entry.objects.filter(
             project=a_summary.project,
             id__in=a_summary.entries_id,
+            lead__confidentiality=Lead.Confidentiality.UNPROTECTED,
```
thenav56, line 60:
```suggestion
            # NOTE: NLP is using LLM, to avoid data leakage we only pass UNPROTECTED
            lead__confidentiality=Lead.Confidentiality.UNPROTECTED,
```

apps/analysis/serializers.py
```diff
@@ -459,9 +460,13 @@ class Meta:
     def validate_analysis_pillar(self, analysis_pillar):
         if analysis_pillar.analysis.project != self.context['request'].active_project:
             raise serializers.ValidationError('Invalid analysis pillar')
+        if self.context['request'].active_project.is_private:
+            raise serializers.ValidationError('Topic model is not allowed for private projects')
         return analysis_pillar
 
     def validate_additional_filters(self, additional_filters):
+        # NOTE: overlapping the lead_confidentialities filter, only public leads are allowed
+        additional_filters['lead_confidentialities'] = [Lead.Confidentiality.UNPROTECTED]
```
thenav56: Let's move this logic to 


apps/analysis/models.py:473
```python3
    def get_entries_qs(self):
        additional_filters = copy.deepcopy(self.additional_filters)
        # NOTE: NLP is using LLM, to avoid data leakage we only pass UNPROTECTED
        additional_filters['lead_confidentialities'] = [Lead.Confidentiality.UNPROTECTED]
        return self._get_entries_qs(self.analysis_pillar, additional_filters)
```

apps/analysis/tests/test_mutations.py
```diff
@@ -392,9 +419,11 @@ def test_automatic_summary(self, trigger_results_mock, RequestHelperMock):
         lead1 = LeadFactory.create(project=self.project)
         lead2 = LeadFactory.create(project=self.project)
         another_lead = LeadFactory.create(project=self.another_project)
+        lead3 = LeadFactory.create(project=self.private_project)
```
thenav56, line 447: Let's create another confidential leads RESTRICTED, CONFIDENTIAL and some entries. Make sure the generated files doesn't include those entries, even if filters allows.

Similar to what we did before here https://github.com/the-deep/server/pull/1502/files#diff-223594694777faf05fef2c6acee1587a281db5d1002813ca7b0e43862d6feda0R256-R268

thenav56 (changes requested):

apps/analysis/tests/test_mutations.py
susilnem: For this one, Let's merge the Mentioned PR and Update the fields validation. They are on same function that's why.

thenav56: @AdityaKhatri Will merge this https://github.com/the-deep/server/pull/1502. After that, let's confirm the above changes.

sudan45 (approved):

